PR #5, opened 2022-06-18
Add Test Session Document Model

mahadzaryab1:
## Notion ticket link

[Create test sessions model](https://www.notion.so/uwblueprintexecs/Create-test-sessions-model-83885eea20e843d48c0cf6ea3ebac2ab)

## Implementation description
* Created a new document model for the `TestSession` collection

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/typescript/models/testSession.model.ts
+++ b/backend/typescript/models/testSession.model.ts
@@ -13,7 +13,12 @@ export interface Result {
    * a list corresponding to the question list with each field indicating
    * the student's answer
    */
-  breakdown: [number];
+  answers: [number];
+  /**
+   * a list corresponding to the question list with each fielding indicating
+   * whether the student got the question right or not
+   * */
+  breakdown: [boolean];
 }
 
 const ResultSchema: Schema = new Schema({
@@ -25,8 +30,12 @@ const ResultSchema: Schema = new Schema({
     type: Number,
     required: true,
   },
-  breakdown: {
+  answers: {
     type: [Number],
+    required: true,
+  },
+  breakdown: {
+    type: [Boolean],
     required: true,
   },
 });
--- a/backend/typescript/models/testSession.model.ts
+++ b/backend/typescript/models/testSession.model.ts
@@ -1,0 +1,53 @@
+import mongoose, { Schema, Document } from "mongoose";
+
+export interface Result {
+  student: string;
+  score: number;
+  breakdown: [boolean];
+}
+
+const ResultSchema: Schema = new Schema({
+  student: {
+    type: String,
+    required: true,
+  },
+  score: {
+    type: Number,
+    required: true,
+  },
+  breakdown: {
+    type: [Boolean],
+    required: true,
+  },
+});
+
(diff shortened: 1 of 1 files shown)


backend/typescript/models/testSession.model.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface Result {
+  /** the name of the student */
+  student: string;
+  /** the score of the student */
+  score: number;
+  /**
+   * a list corresponding to the question list with each fielding indicating
+   * whether the student got the question right or not
+   */
+  breakdown: [boolean];
```
joyce-shi, line 16: So, client has requested that we store student answers to questions (not just whether a question was correct or incorrect). I think by just changing this to `number` instead of `boolean` we should accomplish this? Let me know if this is an oversimplification.

backend/typescript/models/testSession.model.ts
```diff
@@ -0,0 +1,81 @@
...
+    required: true,
+  },
+});
+
+/**
+ * This document contains information about a single test session
+ */
+export interface TestSession extends Document {
+  /** the unique identifier of the test session */
+  id: string;
```
joyce-shi, line 39: Why do we explicitly add a unique identifier here in `TestSession` collection, but not in other collections? i.e `Users`, `Schools`, `Tests`. Is a unique identifier auto-generated, or do we have to add it as a field? If not automatic, maybe we should be adding to other collections as well.

joyce-shi (changes requested):
Same concerns here with using `String` instead of `ObjectId` as reference for `test`, `teacher` and `school` fields, but we can talk about this when we have a chance to sync!

backend/typescript/models/testSession.model.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** the ID of the corresponding test from the Test collection */
+  test: string;
+  /** the ID of the teacher administering the test from the User collection */
+  teacher: string;
+  /** the ID of the school that's administering the test from the School collection */
+  school: string;
+  /** the grade level that is being tested */
+  grade_level: number;
+  /**
```
joyce-shi, line 48: Let's also add a field for `access_code` (the access code that students can use to access the test when it is live) and `start_time` (when test is started). Sorry this wasn't included in the scope of this ticket prior to you working on it, just thought of this!

backend/typescript/models/testSession.model.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface Result {
+  /** the name of the student */
+  student: string;
+  /** the score of the student */
+  score: number;
+  /**
+   * a list corresponding to the question list with each field indicating
+   * the student's answer
+   */
+  breakdown: [number];
```
joyce-shi, line 16: Sorry, I just realized we need both `breakdown` and `answers`. i.e we should store what the student wrote and whether or not the answer is correct.

joyce-shi (changes requested):

joyce-shi (approved):
LGTM! 🛳

PR #10, opened 2022-06-30
Updating Vault post-merge hook to use project path

joyce-shi:
## Implementation description
* Updating post-merge hook to use our project's path `jump-math` (instead of `internal-tools`).


obaranek (approved):
LGTM!

PR #14, opened 2022-07-01
Feat: Create Test Session Service

carissa-tang:
## Notion ticket link

[Create test session](https://www.notion.so/uwblueprintexecs/Create-test-session-71e3938ba11c473abf3dfdbb5c704702)

## Implementation description
* Wrote `createTestSession` function
* Updated `results` to be an optional attribute
* Fixed types of ids in test and testSession models
* Added missing attributes for testSession model
* Wrote one successful test

- *Not sure what, if any, failing case needs to be covered

## Steps to test
1.

* 

## Checklist


backend/testUtils/testSession.ts
```diff
@@ -0,0 +1,9 @@
+export const mockTestSession = {
+    test: "1",
+    teacher: "2",
+    school: "3",
+    grade_level: 4,
+    results: [],
```
carissa-tang, line 6: When we create a test session, would we supply an empty array for `results`? And if that's the case, what would the type for `results` be? Cause right now as `results: [Result]`, it is requiring at least one Result object @joyce-shi

carissa-tang:
Side question... should we be using camelCase?

backend/testUtils/testSession.ts
joyce-shi: yes! we should start the array as empty and then populate it as student results are received. given this, do you mind updating the `Result` schema so the the field is optional? (I think just set `required: false`)

backend/services/interfaces/testSessionService.ts
```diff
@@ -0,0 +1,30 @@
+export interface CreateTestSessionRequestDTO {
```
mahadzaryab1: nit: can we document these interfaces?

backend/services/interfaces/testSessionService.ts
```diff
@@ -0,0 +1,30 @@
...
+  test: string;
+  teacher: string;
+  school: string;
+  grade_level: number;
+  access_code: string;
+  start_time: Date;
+}
+
+export interface CreateTestSessionResponseDTO {
```
mahadzaryab1: same

backend/models/test.model.ts
```diff
@@ -87,7 +87,8 @@ const TestSchema: Schema = new Schema(
       required: true,
     },
     admin: {
-      type: { type: mongoose.Schema.Types.ObjectId, ref: "User" },
+      type: mongoose.Schema.Types.ObjectId,
+      ref: "User",
```
joyce-shi, line 91: thanks for catching this syntax error!

backend/models/testSession.model.ts
```diff
@@ -58,23 +58,28 @@ export interface TestSession extends Document {
...
     teacher: {
-      type: { type: mongoose.Schema.Types.ObjectId, ref: "User" },
+      type: mongoose.Schema.Types.ObjectId,
+      ref: "User",
       required: true,
     },
     school: {
-      type: { type: mongoose.Schema.Types.ObjectId, ref: "School" },
+      type: mongoose.Schema.Types.ObjectId,
+      ref: "School",
```
joyce-shi, line 82: 🥺 ^^ thanks again LOL

mahadzaryab1 (commented):
couple of small comments

joyce-shi (approved):
LGTM, other than what we discussed async. +1 to Mahad's documentation comments, but other than that good to 🛳 !

PR #25, opened 2022-07-10
Implement Method to Get All Test Sessions

mahadzaryab1:
## Notion ticket link

[Get all test sessions](https://www.notion.so/uwblueprintexecs/Get-all-test-sessions-b77ba97e037341ac862deff161b65a53)

## Implementation description
* Added a new function `getAllTestSessions` to fetch all the test sessions from the database
* Fixed some bugs/syntax errors

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/models/testSession.model.ts
+++ b/backend/models/testSession.model.ts
@@ -4,7 +4,7 @@ import mongoose, { Schema, Document } from "mongoose";
  * This interface holds information about the result of a single student
  * on a test
  */
-export interface Result {
+export interface Result extends Document {
   /** the name of the student */
   student: string;
   /** the score of the student */
@@ -13,12 +13,12 @@ export interface Result {
    * a list corresponding to the question list with each field indicating
    * the student's answer
    */
-  answers: [number];
+  answers: number[];
   /**
    * a list corresponding to the question list with each fielding indicating
    * whether the student got the question right or not
    * */
-  breakdown: [boolean];
+  breakdown: boolean[];
 }
 
 const ResultSchema: Schema = new Schema({
@@ -58,7 +58,7 @@ export interface TestSession extends Document {
    * the result of the test session
    * there should be one entry here per student
    * */
-  results?: [Result];
+  results?: Result[];
   /** the code that students can use to access the test when it is live */
   accessCode: string;
   /** the time when the test session is started by teacher */
--- a/backend/services/implementations/__tests__/testSessionService.test.ts
+++ b/backend/services/implementations/__tests__/testSessionService.test.ts
@@ -4,6 +4,7 @@ import db from "../../../testUtils/testDb";
 import MgTestSession from "../../../models/testSession.model";
 import {
   assertResponseMatchesExpected,
+  assertResultsResponseMatchesExpected,
   mockTestSession,
 } from "../../../testUtils/testSession";
 
@@ -30,7 +31,8 @@ describe("mongo testSessionService", (): void => {
     const res = await testSessionService.createTestSession(mockTestSession);
 
     // TODO: uncomment when results are added to test session response object
-    // assertResponseMatchesExpected(mockTestSession, res);
+    assertResponseMatchesExpected(mockTestSession, res);
+    expect(res.results).toBeUndefined();
   });
 
   it("getAllTestSessions", async () => {
@@ -38,5 +40,6 @@ describe("mongo testSessionService", (): void => {
 
     const res = await testSessionService.getAllTestSessions();
     assertResponseMatchesExpected(mockTestSession, res[0]);
+    assertResultsResponseMatchesExpected(mockTestSession, res[0]);
   });
 });
--- a/backend/services/implementations/__tests__/testSessionService.test.ts
+++ b/backend/services/implementations/__tests__/testSessionService.test.ts
@@ -1,7 +1,11 @@
 import TestSessionService from "../testSessionService";
 
-import mockTestSession from "../../../testUtils/testSession";
 import db from "../../../testUtils/testDb";
+import MgTestSession from "../../../models/testSession.model";
+import {
+  assertResponseMatchesExpected,
+  mockTestSession,
+} from "../../../testUtils/testSession";
 
 describe("mongo testSessionService", (): void => {
   let testSessionService: TestSessionService;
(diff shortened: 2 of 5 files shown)


backend/models/testSession.model.ts
```diff
@@ -13,12 +13,12 @@ export interface Result {
    * a list corresponding to the question list with each field indicating
    * the student's answer
    */
-  answers: [number];
+  answers: number[];
```
mahadzaryab1, line 16: this and the following changes fix a syntax error that I made while writing these models

backend/services/implementations/__tests__/testSessionService.test.ts
```diff
@@ -25,10 +29,14 @@ describe("mongo testSessionService", (): void => {
   it("createTestSession", async () => {
     const res = await testSessionService.createTestSession(mockTestSession);
 
-    expect(res.id).not.toBeNull();
-    expect(res).toMatchObject({
-      id: res.id,
-      ...mockTestSession,
-    });
+    // TODO: uncomment when results are added to test session response object
+    // assertResponseMatchesExpected(mockTestSession, res);
```
mahadzaryab1: There's missing logic in `createTestSession` that doesn't map the Results when returning the response object. Should create a separate ticket for it.

backend/testUtils/testSession.ts
```diff
@@ -1,10 +1,49 @@
...
 
-export default mockTestSession;
+export const assertResponseMatchesExpected = (
+  expected: TestSessionRequestDTO,
+  result: TestSessionResponseDTO,
+): void => {
+  expect(result.id).not.toBeNull();
+  expect(result.test.toString()).toEqual(expected.test);
+  expect(result.teacher.toString()).toEqual(expected.teacher);
+  expect(result.school.toString()).toEqual(expected.school);
```
mahadzaryab1, line 31: did we decide on whether we want these to return the actual objects rather than just the IDs? if so, we can create a new ticket for that and do it in a separate PR

backend/services/implementations/__tests__/testSessionService.test.ts
carissa-tang: i omitted the `Results` field because it will be empty when a test session is first created, and was thinking we could add that attribute when a test session is updated. or do you think it would be better to have an empty `Results` field? I was having some trouble figuring out the typing for that

backend/testUtils/testSession.ts
carissa-tang: i believe we are returning the actual objects - i was thinking we could use `getTestById`, `getTeacherById` and `getSchoolById` once they are implemented

carissa-tang (approved):
lgtm! will need to create new tickets to update the mapping of ids to objects and add `result` to response of `createTestSession()`

backend/services/implementations/__tests__/testSessionService.test.ts
mahadzaryab1: ah i see - that's fine! i changed the test up to reflect this

carissa-tang (approved):

backend/models/testSession.model.ts
joyce-shi: nice

backend/testUtils/testSession.ts
```diff
@@ -1,10 +1,54 @@
...
+  ResultRequestDTO,
+  TestSessionRequestDTO,
+  TestSessionResponseDTO,
+} from "../services/interfaces/testSessionService";
+
+export const testResult: ResultRequestDTO = {
+  student: "some-student-name",
+  score: 25,
+  answers: [10, 11],
+  breakdown: [false, true, true],
```
joyce-shi: wait - how can breakdown.length be > than answer.length lol

backend/services/implementations/testSessionService.ts
```diff
@@ -38,6 +38,53 @@ class TestSessionService implements ITestSessionService {
...
+      Logger.error(
+        `Failed to get test sessions. Reason = ${getErrorMessage(error)}`,
+      );
+      throw error;
+    }
+
+    return testSessionDtos;
+  }
+
+  private async mapTestSessionsToTestSessionDTOs(
```
joyce-shi, line 60: not sure if you know the answer to this - but how does mongo handle large requests? i.e say we have thousands of test sessions, if we make a request for "all test sessions", how does it behave?

backend/services/interfaces/testSessionService.ts
```diff
@@ -43,13 +43,49 @@ export interface TestSessionResponseDTO {
...
    * */
-  results?: [Result];
+  results?: ResultResponseDTO[];
   /** the code that students can use to access the test when it is live */
   accessCode: string;
   /** the time when the test session is started by teacher */
   startTime: Date;
 }
 
+export interface ResultRequestDTO {
```
joyce-shi, line 53: why do we add `ResultRequestDTO` and `ResultResponseDTO`? is it for the sake of modularity (since we are using it as a DTO) or is it strictly necessary for another reason?

backend/testUtils/testSession.ts
joyce-shi: @carissa-tang sorry if I'm misunderstanding, but if we end up using those functions, shouldn't we just return Ids instead of objects?

backend/testUtils/testSession.ts
mahadzaryab1: oops I made a typo - fixed now!

backend/services/implementations/testSessionService.ts
mahadzaryab1: think it should be fine - would probably become a problem if we were querying millions of documents

backend/services/interfaces/testSessionService.ts
mahadzaryab1: just to keep the request and response objects independent in case they need to diverge in the future

backend/testUtils/testSession.ts
carissa-tang: @joyce-shi within `createTestSession` we need to validate the 3 given ids and so we were thinking if we are already making the call, we can map the objects to the ids in the response. is there a downside you are considering with this approach?

backend/testUtils/testSession.ts
carissa-tang: i think the validation for this would be within `createTestResult`?

mahadzaryab1: yep exactly @carissa-tang

backend/services/implementations/testSessionService.ts
joyce-shi: cool, sounds good

backend/services/interfaces/testSessionService.ts
joyce-shi: makes sense, good call

backend/testUtils/testSession.ts
joyce-shi: oh shoot, i forgot about this. makes sense! sounds good.

joyce-shi (approved):
LGTM 🛳

PR #30, opened 2022-07-14
Feat: add getTestSessionsByTeacherId

carissa-tang:
## Notion ticket link

[Get test sessions by teacher id](https://www.notion.so/uwblueprintexecs/Get-test-sessions-by-teacher-id-b214d531a91b4a56aecd64288246855d)

## Implementation description
* Updated interface and implementation for `getTestSessionsByTeacherId` which retrieves all TestSessions associated with the given `teacher` id and returns an array of `TestSessionResponseDTO`
* Update implementation and tests for `getTestSessionsByTeacherId` and `getSchoolsBySubregion` to return empty array instead of error when no records match the given query
* Wrote 1 success and 1 failing test case

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -39,22 +39,21 @@ class TestSessionService implements ITestSessionService {
     };
   }
 
-  async deleteTestSession(
-    id: string
-  ): Promise<string> {
-    try{
+  async deleteTestSession(id: string): Promise<string> {
+    try {
       const deletedTestSession = await MgTestSession.findByIdAndDelete(id);
-      if(!deletedTestSession){
+      if (!deletedTestSession) {
         throw new Error(`Test Session id ${id} not found`);
       }
       return id;
-    } catch (error: unknown){
+    } catch (error: unknown) {
       Logger.error(
         `Failed to delete entity. Reason = ${getErrorMessage(error)}`,
       );
       throw error;
     }
   }
+
   async getAllTestSessions(): Promise<Array<TestSessionResponseDTO>> {
     let testSessionDtos: Array<TestSessionResponseDTO> = [];
 
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -6,17 +6,10 @@ import {
 } from "../interfaces/testSessionService";
 import { getErrorMessage } from "../../utilities/errorUtils";
 import logger from "../../utilities/logger";
-import IUserService from "../interfaces/userService";
 
 const Logger = logger(__filename);
 
 class TestSessionService implements ITestSessionService {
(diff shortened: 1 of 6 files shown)


backend/services/implementations/testSessionService.ts
```diff
@@ -57,6 +57,30 @@ class TestSessionService implements ITestSessionService {
...
+  ): Promise<Array<TestSessionResponseDTO>> {
+    let testSessionDtos: Array<TestSession> | null;
+
+    try {
+      testSessionDtos = await MgTestSession.find({
+        teacher: { $eq: teacherId },
+      });
+
+      // check if no testSessions are associated with the given teacherId
+      if (!testSessionDtos || !testSessionDtos.length) {
```
mahadzaryab1: why can't we just check if the length == 0?

mahadzaryab1: now that I think about this some more - does it make sense to throw an error here. Why not just return an empty array?

carissa-tang: hm i think your point makes sense to return an empty array - will update the code!

backend/services/implementations/testSessionService.ts
```diff
@@ -57,6 +57,27 @@ class TestSessionService implements ITestSessionService {
     return testSessionDtos;
   }
 
+  async getTestSessionsByTeacherId(
+    teacherId: string,
+  ): Promise<Array<TestSessionResponseDTO>> {
+    let testSessionDtos: Array<TestSessionResponseDTO> = [];
+    let field = "teacher";
```
carissa-tang: I was wondering if it made sense to combine `getTestSessionsByTeacherId`, `getTestSessionsBySchoolId` and `getTestSessionsByTestId` and add a second param for the `field` which is either "teacher", "school" or "test" -- since the code for all three is the same. or is it better to keep them separate?

carissa-tang: also they're all done already so idk if refactoring is too extra lol

joyce-shi: hmmm. i think either, or is probably fine. the way it is now probably incurs a bit of code repetition, but makes the code a bit easier to skim at first glance i think. imo, not worth the extra effort to refactor, but if you feel strongly about this we can discuss!

joyce-shi: thanks for thinking about this though!! i love that you're always thinking about ways to improve our codebase (:

joyce-shi (commented):
+1 to Mahad's comment, other than that looks good! lmk when you update & ill approve (:

carissa-tang:
> 

addressed all comments!

mahadzaryab1 (approved):
lgtm!

PR #37, opened 2022-07-28
setup raw color schema for design library

joyce-shi:
## Notion ticket link

[Raw Colors](https://www.notion.so/uwblueprintexecs/Raw-Colours-4f277d847a4441e09bafacefcaa71182)

## Implementation description
* Added files necessary to add custom theme components.
* Extended base Chakra theme to include custom colours, as per [design file](https://www.figma.com/file/Uia7C5xwqDlkLnrXMyt5hn/Design-System?node-id=95%3A3399).
* Added `<DesignLibraryExamples>` to landing page of app. As we create custom components/themes/styles we can include instances of them here for testing purposes. It's not very pretty, but get's the job done! An example of how to use custom colours is included as well.

Screen Shot 2022-07-28 at 5 17 12 PM

**Note**: for now, colours can be accessed by using the following format: `blue.50`, where the magnitude of the number correlates with the darkness of the shade. This aligns with Chakra UI's standard. There is a future ticket to investigate whether aliasing is possible (for example, refer to a color as `jump-math-very-light-blue`) .

## Steps to test
Run `yarn start` in `frontend` folder. Changes can be seen on `localhost:3000`.

## Checklist


frontend/src/components/pages/Default.tsx
```diff
@@ -37,6 +38,21 @@ const TeamInfoDisplay = () => {
   );
 };
 
+const DesignLibraryExamples = () => {
+  const { teamName, numTerms, members, isActive } = useContext(SampleContext);
+  return (
+    <div>
+      {/* Example: how to apply custom colors */}
+      <ButtonGroup gap='4'>
+          <ChakraButton bg="blue.50" color="white"> Test Chakra Button </ChakraButton>
```
carissa-tang: I'm just being unnecessarily picky, but maybe we can write the specific colour description instead of "Test Chakra Button" (i.e. "blue", "medium blue", "light blue", "very light blue")

frontend/src/themes/colors.ts
```diff
@@ -0,0 +1,35 @@
...
+    grey: {
+      50: '#FFFFFF',
+      100: '#F4F4F4',
+      200: '#BEBEBE',
+      300: '#636363',
+      400: '#272727',
+    }
+};
+
+export default colors;
```
mahadzaryab1: missing new line here

frontend/src/themes/index.ts
```diff
@@ -0,0 +1,8 @@
+import { extendTheme, theme as base } from "@chakra-ui/react";
+import colors from "./colors"
+
+const jumpMathTheme = extendTheme({
+  colors,
+});
+
+export default jumpMathTheme;
```
mahadzaryab1: same

frontend/src/themes/index.ts
```diff
@@ -0,0 +1,8 @@
+import { extendTheme, theme as base } from "@chakra-ui/react";
+import colors from "./colors"
+
+const jumpMathTheme = extendTheme({
```
mahadzaryab1: nit: can just leave this generic and say `theme`

frontend/src/components/pages/Default.tsx
```diff
@@ -37,6 +38,21 @@ const TeamInfoDisplay = () => {
   );
 };
 
+const DesignLibraryExamples = () => {
+  const { teamName, numTerms, members, isActive } = useContext(SampleContext);
```
carissa-tang:
```suggestion
```
can remove since context values aren't being used

carissa-tang (changes requested):

frontend/src/components/pages/Default.tsx
joyce-shi: oops lol missed that, ty!

frontend/src/components/pages/Default.tsx
joyce-shi: no, love it! good catch

mahadzaryab1 (commented):
couple of small comments - also, I believe we're supposed to be using `yarn` right? If that's the case - I believe we should only be committing `yarn.lock` and not `package-lock.json`.

joyce-shi:
> couple of small comments - also, I believe we're supposed to be using `yarn` right? If that's the case - I believe we should only be committing `yarn.lock` and not `package-lock.json`.

oops! good catch, i keep forgetting and using npm out of habit tbh

carissa-tang (approved):
lgtm! gonna assume hex codes are correct LOL

joyce-shi:
> 

LOL if not someone will find out :D

PR #40, opened 2022-07-29
Feat: Compute test grades

carissa-tang:
## Notion ticket link

[Compute test grades](https://www.notion.so/uwblueprintexecs/Compute-test-grades-4c7b0c281b7b4e02be33b55bd500a5c5)

## Implementation description
* computeTestGrades takes in a `UngradedResultDTO` and its associated test id, and returns a `GradedResultResponseDTO` with the calculated score and breakdown
* Will be called within `updateTestSession` > `createTestResult`
   * Can look here https://github.com/uwblueprint/jump-math/pull/36/files for full picture of how this function intends to be used
* Wrote tests
   * Successful case
   * Invalid test id 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/models/testSession.model.ts
+++ b/backend/models/testSession.model.ts
@@ -1,6 +1,14 @@
 import mongoose, { Schema, Document } from "mongoose";
 
 /**
+ * An enum containing the grading status of a Result
+ */
+ export enum GradingStatus {
+  GRADED,
+  UNGRADED,
+}
+
+/**
  * This interface holds information about the result of a single student
  * on a test
  */
@@ -20,6 +28,8 @@ export interface Result extends Document {
    * whether the student got the question right or not
    * */
   breakdown: boolean[];
+  /** the grading status of the result */
+  gradingStatus: GradingStatus;
 }
 
 const ResultSchema: Schema = new Schema({
@@ -39,6 +49,12 @@ const ResultSchema: Schema = new Schema({
     type: [Boolean],
     required: true,
   },
+  gradingStatus: {
+    type: String,
+    required: true,
+    default: GradingStatus.UNGRADED,
+    enum: Object.keys(GradingStatus),
+  }
 });
 
 /**
--- a/backend/models/testSession.model.ts
+++ b/backend/models/testSession.model.ts
@@ -11,7 +11,7 @@ export interface Result extends Document {
   score: number;
   /**
    * a list corresponding to the question list with each field indicating
-   * the student's answer - either the numeric answer (for short answer) 
+   * the student's answer - either the numeric answer (for short answer)
    * or the option's corresponding index (for multiple choice)
    */
   answers: number[];
--- a/backend/models/testSession.model.ts
+++ b/backend/models/testSession.model.ts
@@ -11,7 +11,8 @@ export interface Result extends Document {
   score: number;
   /**
    * a list corresponding to the question list with each field indicating
-   * the student's answer
+   * the student's answer - either the numeric answer (for short answer) 
+   * or the option's corresponding index (for multiple choice)
    */
   answers: number[];
   /**
--- a/backend/models/testSession.model.ts
+++ b/backend/models/testSession.model.ts
@@ -3,7 +3,7 @@ import mongoose, { Schema, Document } from "mongoose";
 /**
  * An enum containing the grading status of a Result
  */
- export enum GradingStatus {
+export enum GradingStatus {
   GRADED,
   UNGRADED,
 }
@@ -54,7 +54,7 @@ const ResultSchema: Schema = new Schema({
     required: true,
     default: GradingStatus.UNGRADED,
     enum: Object.keys(GradingStatus),
-  }
+  },
 });
 
 /**
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -1,16 +1,26 @@
 import MgTestSession, { TestSession } from "../../models/testSession.model";
 import {
   ITestSessionService,
+  NewResultDTO,
+  ResultDTO,
   TestSessionRequestDTO,
   TestSessionResponseDTO,
 } from "../interfaces/testSessionService";
 import { getErrorMessage } from "../../utilities/errorUtils";
 import logger from "../../utilities/logger";
+import { MultipleChoiceMetadata, NumericQuestionMetadata, Question, QuestionType } from "../../models/test.model";
+import { ITestService, TestResponseDTO } from "../interfaces/testService";
 
 const Logger = logger(__filename);
 
 class TestSessionService implements ITestSessionService {
   /* eslint-disable class-methods-use-this */
+  testService: ITestService;
+
+  constructor(testService: ITestService) {
+    this.testService = testService;
+  }
+
   async createTestSession(
     testSession: TestSessionRequestDTO,
   ): Promise<TestSessionResponseDTO> {
@@ -186,6 +196,69 @@ class TestSessionService implements ITestSessionService {
     }
 
     return testSessionDtos;
+  }
+
+  /*
+  * computeTestGrades computes the breakdown and score of a given result and returns the ResultDTO
+  */
+  async computeTestGrades(result: NewResultDTO, testId: string): Promise<ResultDTO> {
+    let resultDto: ResultDTO;
+    let questionsCorrect: number = 0; // the number of questions the student gets correct
+    let studentAnswers: number[] = result.answers; // the array of the student's raw answers
+    let updatedScore: number = 0.00; // the final score that will be returned
+    let updatedBreakdown: boolean[] = []; // the final breakdown that will be returned
+
+    try {
+      const test: TestResponseDTO = await this.testService.getTestById(testId);
+
+      // check if the number of test questions matches the number of student answers
+      if (test.questions.length != studentAnswers.length) {
+        throw new Error("One or more of the student's test answers was not found");
+      }
+
+      // check the correctness of each question and answer pair
+      test.questions.forEach((question: Question, i) => {
+        let actualAnswer: number;
+
+        if (question.questionType == QuestionType.MULTIPLE_CHOICE) {
+          // set the actualAnswer for the multiple choice question
+          const questionMetadata = question.questionMetadata as MultipleChoiceMetadata;
+          actualAnswer = Number(questionMetadata.options[questionMetadata.answerIndex]);
+        } else {
+          // set the actualAnswer for the numeric answer question
+          const questionMetadata = question.questionMetadata as NumericQuestionMetadata;
+          actualAnswer = questionMetadata.answer;
+        }
+
+        // check if the student's answer and the actual answer match
+        if (studentAnswers[i] == actualAnswer) {
+          questionsCorrect += 1; // update the counter
+          updatedBreakdown[i] = true;
+        } else {
+          updatedBreakdown[i] = false;
+        }
+      })
+
+      // compute student's score as a percentage to two decimal places (e.g. 1/3 => 33.33)
+      updatedScore = parseFloat((questionsCorrect * 100 / studentAnswers.length).toFixed(2));
+
+      resultDto = {
+        student: result.student,
+        score: updatedScore,
+        answers: result.answers,
+        breakdown: updatedBreakdown,
+      }
+    } catch (error: unknown) {
+      Logger.error(
+        `Failed to compute test grades for result=${result}. Reason = ${getErrorMessage(
+          error,
+        )}`,
+      );
+      throw error;
+    }
+
+    // return Result with updated score and breakdown values
+    return resultDto;
   }
 }
 
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -8,7 +8,12 @@ import {
 } from "../interfaces/testSessionService";
 import { getErrorMessage } from "../../utilities/errorUtils";
 import logger from "../../utilities/logger";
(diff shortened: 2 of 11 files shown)


backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +197,69 @@ class TestSessionService implements ITestSessionService {
...
+    let questionsCorrect: number = 0; // the number of questions the student gets correct
+    let studentAnswers: number[] = result.answers; // the array of the student's raw answers
+    let updatedScore: number = 0.00; // the final score that will be returned
+    let updatedBreakdown: boolean[] = []; // the final breakdown that will be returned
+
+    try {
+      const test: TestResponseDTO = await this.testService.getTestById(testId);
+
+      // check if the number of test questions matches the number of student answers
+      if (test.questions.length != studentAnswers.length) {
```
carissa-tang: I don't know if this will even be a possible case, but I have it here anyways

Side note: Is it possible for a student to leave a question blank? If so, should we check if the answer=null or smth @joyce-shi

joyce-shi: Hmm,  I don't think so. If a student left the answer blank, I think best practice would be to populate from the frontend with a `null` value. So, given this, I'm not sure if `test.questions.length` would ever vary from the `studentAnswers.length`? Let me know if I'm missing a case here!

joyce-shi:
Side Note: interesting how `yarn fix` seemed to make changes... maybe a good example for why we should have automatic pre-commit linting... I feel like those sort of changes make it harder to review PRs

backend/services/interfaces/testSessionService.ts
```diff
@@ -41,33 +41,24 @@ export interface TestSessionResponseDTO {
    * the result of the test session
    * there should be one entry here per student
    * */
-  results?: ResultResponseDTO[];
+  results?: ResultDTO[];
```
joyce-shi: Hmmm... so I think we had some context for why we had kept Response and Request DTOs separate (despite their similarities), see [here](https://github.com/uwblueprint/jump-math/pull/25#discussion_r917479900). I think Mahad makes a good point about keeping them separate for the sake of modularity!

If you're interested, you can read a bit more about this [here](https://stackoverflow.com/questions/44349199/reusing-dto-for-various-request-response-types-vs-explicitness-of-what-is-requir)! (Although, the link talks about separating common elements into some sort of base class, but I think that's overkill for our use case here)

backend/services/interfaces/testSessionService.ts
```diff
@@ -41,33 +41,24 @@ export interface TestSessionResponseDTO {
...
   /** the code that students can use to access the test when it is live */
   accessCode: string;
   /** the time when the test session is started by teacher */
   startTime: Date;
 }
 
-export interface ResultRequestDTO {
+export interface NewResultDTO {
```
joyce-shi: What are your thoughts on calling this something like `UnscoredResultDTO` or `UnmarkedResultDTO` or something? To include a bit more context in naming & make it a bit more intuitive/descriptive. Or something else if you can think of a better name than those LOL.

joyce-shi: Let's also add some documentation for the interface, explaining why its necessary as well (i.e what we talked about on call last work session).

backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +204,78 @@ class TestSessionService implements ITestSessionService {
...
   }
+
+  /*
+   * computeTestGrades computes the breakdown and score of a given result and returns the ResultDTO
+   */
+  async computeTestGrades(
+    result: NewResultDTO,
+    testId: string,
+  ): Promise<ResultDTO> {
+    let resultDto: ResultDTO;
```
joyce-shi: Re: similar to the comment made in `interfaces/testSessionService.ts` let's call this `markedResult` or `gradedResult` or something along those lines.

backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +204,78 @@ class TestSessionService implements ITestSessionService {
...
+    let questionsCorrect = 0; // the number of questions the student gets correct
+    const studentAnswers: number[] = result.answers; // the array of the student's raw answers
```
joyce-shi: baaaaa... I think when we defined models, we said that this would always be of type `number`, and if it was short answer it would be the numeric answer and if it was multiple choice, it would be the index. What are thoughts on adding a quick comment here that this is the case? Because I think this logic is not super intuitive.

And also updating the `testSession.model.ts` to say something to that effect as well!

backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +204,78 @@ class TestSessionService implements ITestSessionService {
...
+    let updatedScore = 0.0; // the final score that will be returned
+    const updatedBreakdown: boolean[] = []; // the final breakdown that will be returned
+
+    try {
+      const test: TestResponseDTO = await this.testService.getTestById(testId);
+
+      // check if the number of test questions matches the number of student answers
+      if (test.questions.length !== studentAnswers.length) {
```
joyce-shi: Not sure if this check is necessary, as discussed [here](https://github.com/uwblueprint/jump-math/pull/40#discussion_r932817119). Let me know if I'm missing something!

backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +204,78 @@ class TestSessionService implements ITestSessionService {
...
+   * computeTestGrades computes the breakdown and score of a given result and returns the ResultDTO
+   */
+  async computeTestGrades(
+    result: NewResultDTO,
+    testId: string,
+  ): Promise<ResultDTO> {
+    let resultDto: ResultDTO;
+    let questionsCorrect = 0; // the number of questions the student gets correct
+    const studentAnswers: number[] = result.answers; // the array of the student's raw answers
+    let updatedScore = 0.0; // the final score that will be returned
```
joyce-shi: nit: let's call this `computedScore` or something, since I don't think we have a score before, so don't think we are updating anything?

backend/services/implementations/testSessionService.ts
```diff
@@ -187,6 +204,78 @@ class TestSessionService implements ITestSessionService {
...
+    const updatedBreakdown: boolean[] = []; // the final breakdown that will be returned
```
joyce-shi: same as above ^^.

backend/services/implementations/testSessionService.ts
carissa-tang: Wait so, on backend would we store smth like `studentAnswers=[10, null, 20]` if second question is blank

joyce-shi (changes requested):
Left a couple comments! Mostly around code readability and variable naming. Let me know if you have thoughts please!

Also, general comment about this PR: I noticed that there are a lot of comments throughout the body of the `computeTestGrades` function. While I do think we've been adding a lot of comments in this project in general, I think up until now it's been okay because code comments are generally good practice for function and interface declarations. This is because they allow for future developers to skim through code and see what functions do without spending too much time in the weeds of the code. (More info [here](https://codingart.readthedocs.io/en/latest/c/Comments.html#:~:text=Every%20function%20declaration%20should%20have,comments%20in%20the%20function%20definition.)!)

But in the body of the function, I think general consensus is that there should only be code comments when there is a non-intuitive case (and there is no way to make it intuitive, i.e the fact that we store all `studentAnswers` as numbers). So for things like "updating the counter", there should be no need to add a comment! If it's not clear that we are incrementing the counter, maybe it is helpful to rename the variable, i.e include the term counter, or something.

Some best practices for commenting can be found [here](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/)! If you're interested in learning more. I also think that when it comes to this sort of code styling / readability commentary, it gets pretty subjective, so use your best judgement and of course ping me if you want to discuss further / have questions!

(Also, sorry for such a long monologue LOL, I hope this is helpful and interesting!)

But also, 🔥 work overall! I think you've been working on two of the more complex tickets out of the backend work we've been doing and I'm impressed with your clarity and comprehensiveness of thinking around how the code should function. (:

backend/services/implementations/testSessionService.ts
joyce-shi: I think so. Let me know if you think there is a better alternative! And if we do this, we should have `null` checks in the `computeGrades` function then as well, I think.

backend/testUtils/testSession.ts
```diff
@@ -1,22 +1,78 @@
...
+  ],
+  grade: 11,
+};
+
+/**
+ * Mock Test Results
+ */
+export const mockUngradedTestResult: UngradedResultDTO = {
+  student: "some-student-name",
+  answers: [10, 11, 1, null],
```
carissa-tang, line 57: I've updated the result interfaces to `(number | null)[]` to allow us to store missing answers as null - lmk if this makes sense or if there's a better implementation / or if we are even allowing for blank answers in the first place? (can just make it required field in FE)

backend/services/implementations/testSessionService.ts
carissa-tang: i've updated the interfaces! should work with the current logic since we directly check for strict equality `studentAnswers[i] === actualAnswer`

backend/services/interfaces/testSessionService.ts
```diff
@@ -41,14 +41,28 @@ export interface TestSessionResponseDTO {
...
   /** the time when the test session is started by teacher */
   startTime: Date;
 }
 
-export interface ResultRequestDTO {
+/**
+ * This interface contains the request object that is fed into the test session
+ * service that represents a new test result that has not yet been graded
+ */
+export interface UngradedResultDTO {
```
mahadzaryab1: i'm wondering why we need to separate result objects? can we not just have one with the `breakdown` and `score` set to null or empty?

mahadzaryab1: we could potentially even add a `status` field and have it default to `ungraded` and then set it to `graded` once it is.

backend/testUtils/testSession.ts
```diff
@@ -1,22 +1,78 @@
...
   TestSessionRequestDTO,
   TestSessionResponseDTO,
 } from "../services/interfaces/testSessionService";
+import { QuestionType } from "../models/test.model";
 
-export const testResult: ResultRequestDTO = {
+/**
+ * Mock Tests
+ */
+export const mockTest = {
```
mahadzaryab1: should this go in `testUtils/test.ts`?

backend/services/implementations/testSessionService.ts
```diff
@@ -189,6 +204,69 @@ class TestSessionService implements ITestSessionService {
...
+    try {
+      const test: TestResponseDTO = await this.testService.getTestById(testId);
+
+      test.questions.forEach((question: Question, i) => {
+        let actualAnswer: number;
+
+        if (question.questionType === QuestionType.MULTIPLE_CHOICE) {
+          const questionMetadata = question.questionMetadata as MultipleChoiceMetadata;
+          actualAnswer = questionMetadata.answerIndex;
+        } else {
```
mahadzaryab1: can we switch this to an `else if`? just in case we decide to introduce another question type. Another option is to use a switch statement

backend/services/implementations/testSessionService.ts
```diff
@@ -189,6 +204,69 @@ class TestSessionService implements ITestSessionService {
...
+          const questionMetadata = question.questionMetadata as NumericQuestionMetadata;
+          actualAnswer = questionMetadata.answer;
+        }
```
mahadzaryab1: we can move this entire block into a helper function called `getCorrectAnswer` or something

backend/testUtils/testSession.ts
```diff
@@ -1,22 +1,78 @@
...
+  id: "62c248c0f79d6c3c9ebbea95",
+  name: "test",
+  duration: 300,
+  admin: "62c248c0f79d6c3c9ebbea94",
+  questions: [
+    {
+      questionType: QuestionType.NUMERIC_ANSWER,
+      questionPrompt: "Numeric answer question",
+      questionMetadata: {
+        answer: 10,
```
mahadzaryab1: let's have one of these as a float to ensure it holds there

mahadzaryab1 (commented):
nice work - added a few comments

backend/models/testSession.model.ts
```diff
@@ -11,7 +11,8 @@ export interface Result extends Document {
   score: number;
   /**
    * a list corresponding to the question list with each field indicating
-   * the student's answer
+   * the student's answer - either the numeric answer (for short answer)
```
joyce-shi, line 22: beauty ty 🔥

joyce-shi (approved):
LGTM. I think Mahad has a few comments outstanding, but once you address those you can merge!

backend/services/interfaces/testSessionService.ts
carissa-tang: i like this suggestion! @joyce-shi what do you think

backend/services/implementations/testSessionService.ts
carissa-tang: switch won't work cause `QuestionType.MULTIPLE_CHOICE=0` always evaluates to true :(. i wonder if there's a way to make it work... but will use if statement for simplicity

mahadzaryab1: I think it's possible - see [here](https://stackoverflow.com/questions/27747437/typescript-enum-switch-not-working). I'll leave it up to you if you want to change it

mahadzaryab1 (approved):
lgtm! great work

PR #45, opened 2022-08-05
Feat: Add buttons theme

carissa-tang:
## Notion ticket link

[Buttons](https://www.notion.so/uwblueprintexecs/Buttons-397f43bf1206417f812ed7563f2c699b)

## Implementation description
* Added `primary`, `secondary` and `tertiary` buttons for the following states
   * default
   * pressed
   * hover 
   * disabled

## Steps to test

https://user-images.githubusercontent.com/69697180/183558116-229a2406-01d0-4156-b946-55edd14610a7.mov

* 

## Checklist

### DIFF
--- a/frontend/src/components/pages/Default.tsx
+++ b/frontend/src/components/pages/Default.tsx
@@ -48,11 +48,6 @@ const DesignLibraryExamples = () => {
           <ChakraButton variant="tertiary">Tertiary</ChakraButton>
       </ButtonGroup>
       <ButtonGroup gap='4'>
-          <ChakraButton variant="primary">Primary</ChakraButton>
-          <ChakraButton variant="secondary">Secondary</ChakraButton>
-          <ChakraButton variant="tertiary">Tertiary</ChakraButton>
-      </ButtonGroup>
-      <ButtonGroup gap='4'>
           <ChakraButton variant="primary" isActive>Primary</ChakraButton>
           <ChakraButton variant="secondary" isActive>Secondary</ChakraButton>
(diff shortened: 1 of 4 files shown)


frontend/src/components/pages/Default.tsx
```diff
@@ -38,17 +38,26 @@ const TeamInfoDisplay = () => {
...
 };
 
+// Example: how to apply custom colors
 const DesignLibraryExamples = () => {
   return (
-    <div>
-      {/* Example: how to apply custom colors */}
+    <VStack spacing={4} bg="yellow.50" padding={2}>
       <ButtonGroup gap='4'>
-          <ChakraButton bg="blue.50" color="white"> Very Light Blue Button </ChakraButton>
```
carissa-tang, line 46: i replaced the original buttons because i think we will probably only use button variants moving forward? (rather than directly styling a button) but lmk if you still want to keep it as an example

joyce-shi: makes sense to me!

joyce-shi (commented):
Overall, LGTM! Some quick comments: 
(1) I noticed that design has created buttons with icons, can we handle this case here as well? I know that we haven't worked on importing an icon pack yet, but I think it's just a one-line import, so I assigned you that ticket as well (see: https://www.notion.so/uwblueprintexecs/Iconography-1b2aac290cd0471e94dacb95b77c3d0e).
(2) Do we use the correct font here? Or do buttons automatically use the default text font? I know we haven't worked on implementing typography here yet, so maybe the action item is just to leave a TODO comment.

joyce-shi (approved):

PR #47, opened 2022-08-09
Feat: add statisticsService

carissa-tang:
## Notion ticket link

[Implement functionality to query for school averages](https://www.notion.so/uwblueprintexecs/Implement-functionality-to-query-for-school-averages-08591015e61f4156a54ebda26f28c009)

## Implementation description
* Created statisticsService - implementation and interface files
* Add StatisticsResponseDTO
* Add `getAverageScoresBySchool` with passing and failing tests

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/services/implementations/statisticsService.ts
+++ b/backend/services/implementations/statisticsService.ts
@@ -6,7 +6,7 @@ import { ISchoolService, SchoolResponseDTO } from "../interfaces/schoolService";
 import {
   IStatisticsService,
   AggregateTestSessionDTO,
-  StatisticsResponseDTO,
+  StatisticsBySchoolDTO,
 } from "../interfaces/statisticsService";
 import { ITestService, TestResponseDTO } from "../interfaces/testService";
 import { ITestSessionService } from "../interfaces/testSessionService";
@@ -38,11 +38,11 @@ class StatisticsService implements IStatisticsService {
 
   async getAverageScoresBySchool(
     testId: string,
-  ): Promise<Array<StatisticsResponseDTO>> {
-    let statisticsDTOs: Array<StatisticsResponseDTO> = [];
+  ): Promise<Array<StatisticsBySchoolDTO>> {
+    let statisticsBySchoolDtos: Array<StatisticsBySchoolDTO> = [];
 
     try {
-      const statistics: AggregateTestSessionDTO[] = await MgTestSession.aggregate(
+      const statisticsBySchool: AggregateTestSessionDTO[] = await MgTestSession.aggregate(
         [
           // Stage 1: Filter Test Session documents by test id
           {
@@ -56,14 +56,8 @@ class StatisticsService implements IStatisticsService {
             $group: {
               _id: "$school",
               averageScore: { $avg: "$results.score" },
-              resultsPerTestArr: {
-                $push: {
-                  $map: {
-                    input: "$results.breakdown",
-                    as: "result",
-                    in: { $cond: ["$$result", 1, 0] },
-                  },
-                },
+              resultBreakdowns: {
+                $push: "$results.breakdown",
               },
             },
           },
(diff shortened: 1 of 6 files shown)


backend/services/implementations/statisticsService.ts
```diff
@@ -0,0 +1,147 @@
...
+          error,
+        )}`,
+      );
+      throw error;
+    }
+
+    return statisticsDTOs;
+  }
+
+  private computeAverageScorePerQuestion(
```
carissa-tang, line 94: would prefer doing this directly in the aggregation pipeline but haven't found a good way to implement :(.  the closest solution would be using `$zip` but there's this [bug](https://jira.mongodb.org/browse/SERVER-27964?page=com.atlassian.streams.streams-jira-plugin%3Aactivity-stream-issue-tab) that restricts the function's use to arrays with fixed dimensions.

carissa-tang: essentially we want to transpose an array of test results (i.e. `[result1, result2, result3,...]` with each result being an array of student answers) to `[question1_answers, question2_answers, ...]` with each entry being an array of all answers for question `i`

carissa-tang:
i should also refactor some of my code since i think some of it can be reused for the other pipelines - but will hold off until i get an idea of how the other ones will be implemented

backend/services/interfaces/statisticsService.ts
```diff
@@ -0,0 +1,36 @@
...
+  id: string;
+  /** the average overall test score */
+  averageScore: number;
+  /** the array of test result breakdowns */
+  resultsPerTestArr: number[][];
+  /** the array of average test scores for each Question */
+  averageScoresByQuestions: number[];
+}
+
+export interface StatisticsResponseDTO {
```
joyce-shi: Should we name this something more specific? Since we will have this type of DTO for every category that we breakdown by (i.e school, sub-region, country)

backend/services/implementations/statisticsService.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    return averageResultsPerQuestion;
+  }
+
+  private async mapStatisticsToStatisticsDTOs(
+    testId: string,
+    statistics: AggregateTestSessionDTO[],
+  ): Promise<Array<StatisticsResponseDTO>> {
+    const testSessionDtos: Array<StatisticsResponseDTO> = await Promise.all(
```
joyce-shi: Not sure if I'm just brainfreezing, but why is this called `testSessionDtos`?

backend/services/implementations/statisticsService.ts
```diff
@@ -0,0 +1,147 @@
...
+          },
+          // Stage 2: Group remaining documents by school id and add aggregate fields
+          {
+            $unwind: "$results",
+          },
+          {
+            $group: {
+              _id: "$school",
+              averageScore: { $avg: "$results.score" },
+              resultsPerTestArr: {
```
joyce-shi: Are we mapping boolean values -> 0/1 binary values here? If so, would it be more intuitive to leave this as `true/false` values here and then do a temporary conversion in `computeAverageScorePerQuestion`  instead? To avoid storing magic numbers.

backend/services/interfaces/statisticsService.ts
```diff
@@ -0,0 +1,36 @@
...
+  /** the corresponding test from the Test collection */
+  test: TestResponseDTO;
```
joyce-shi: Do we need to store `TestResponseDTO` for each `SchoolResponseDTO`? 
i.e maybe it makes more sense to store `school: Array<SchoolResponseDTO>` instead of `SchoolResponseDTO`? Let me know your thoughts here!
maybe then, it would also eliminate the need for `mapStatisticsToStatisticsDTOs` helper?

joyce-shi (changes requested):
Left a couple comments about about structure of the changes. Let me know if I'm missing the point of your PR, not sure if I'm completely understanding all aspects!

backend/services/implementations/statisticsService.ts
joyce-shi: I think it's okay to leave out of aggregation pipeline, as long as the meaning is clear! Which I think it is here.

joyce-shi:
> Left a couple comments about about structure of the changes. Let me know if I'm missing the point of your PR, not sure if I'm completely understanding all aspects!

Tagging @mahadzaryab1 for a second pair of eyes.

backend/services/interfaces/statisticsService.ts
carissa-tang: yes i was thinking that.. also wanted to see how others would implement (e.g. we could have a "filter/grouping" field instead that takes in either a school, sub-region or country). but might be more clear to keep them separate.

backend/services/implementations/statisticsService.ts
carissa-tang: LOL forgot to rename - thanks for catching that!

backend/services/interfaces/statisticsService.ts
carissa-tang: so would we return `Array<StatisticsResponseDTO>` or `Array<SchoolResponseDTO>` ? i'm a bit confused by what you mean

backend/services/interfaces/statisticsService.ts
carissa-tang: will make separate one for now! can refactor later if needed

carissa-tang: actually, i'll refactor AggregateTestSessionDTO to be reusable and we can have separate response dtos

PR #50, opened 2022-08-24
Brian/add typography

brian386:
## Notion ticket link

[Typography](https://www.notion.so/uwblueprintexecs/47bf4727a2444bccad4a91952898b8f3?v=9658436bf9be409e90581631a7155f16&p=96722d03ac204b4d8aeaa432b829a605&pm=s)

## Implementation description
* 

## Steps to test
1. Added some examples of different types of text - check if they look good

* 

## Checklist


joyce-shi (approved):
i think we should be omitting the `package-lock.json` here! other than that lgtm 🚀

frontend/src/components/pages/Default.tsx
```diff
@@ -18,6 +20,43 @@ const Button = ({ text, path }: ButtonProps) => {
...
+      <div>Active: {isActive ? "Yes" : "No"}</div>
+    </div>
+  );
+};
+
+// Example: how to apply custom colors
+const DesignLibraryExamples = () => {
+  return (
+    <>
+      <VStack>
```
carissa-tang: can we move this to our new file called `ComponentLibrary.tsx`?

frontend/src/components/pages/Default.tsx
```diff
@@ -18,6 +20,43 @@ const Button = ({ text, path }: ButtonProps) => {
   );
 };
 
+const TeamInfoDisplay = () => {
```
carissa-tang: this is starter code that we can remove!

carissa-tang (changes requested):

frontend/src/components/pages/Default.tsx
```diff
@@ -18,6 +20,43 @@ const Button = ({ text, path }: ButtonProps) => {
...
+    </div>
+  );
+};
+
+// Example: how to apply custom colors
+const DesignLibraryExamples = () => {
+  return (
+    <>
+      <VStack>
+        <Box textStyle='header1'>header1</Box>
```
carissa-tang: also nit: i think it might be better to use `<Text>` instead of `<Box>` in the examples (just a little more fitting, but they both basically do the same thing)

frontend/src/components/pages/Default.tsx
joyce-shi: oh +1 good catch

carissa-tang (approved):
lgtm!

PR #54, opened 2022-10-02
feat: get users by role and tests

victoranyun:
## Notion ticket link

[Get Users By Role](https://www.notion.so/uwblueprintexecs/Get-Users-by-Role-a1b29d4a578e4fe4a2adb191e66310ef)

## Implementation description
* Added new query to get users by role
* Added tests to verify correctness

## Steps to test
1. yarn test

* Not sure if we need the graphql endpoint, can remove

## Checklist

### DIFF
--- a/backend/graphql/resolvers/userResolvers.ts
+++ b/backend/graphql/resolvers/userResolvers.ts
@@ -29,6 +29,12 @@ const userResolvers = {
     users: async (): Promise<UserDTO[]> => {
       return userService.getUsers();
     },
+    usersByRole: async (
+      _parent: undefined,
+      { role }: { role: string },
+    ): Promise<UserDTO[]> => {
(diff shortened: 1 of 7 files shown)


backend/graphql/resolvers/userResolvers.ts
```diff
@@ -29,6 +29,12 @@ const userResolvers = {
     users: async (): Promise<UserDTO[]> => {
       return userService.getUsers();
     },
+    usersByRole: async (
+      _parent: undefined,
```
joyce-shi, line 33: looks right - but for my own understanding, what is the `parent` argument?

joyce-shi (approved):
overall, lgtm. but not sure what you mean by "not sure if we need the graphql endpoint", could you be specific about which part is optional!

PR #59, opened 2022-10-13
feat: user redirect

chenjanet:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/User-Redirect-fb13debc0e12438481b0bd3ff764e52c)

## Implementation description
* `Login.tsx`: added logic to redirect user to the correct page (`TeacherPage` or `AdminPage`) depending on role
* `TeacherPage.tsx`, `AdminPage.tsx`: added buttons for logging out, changing password, etc. as described in the Notion ticket
* `Default.tsx`: deleted due to redundancy

## Steps to test
1. Log in as an admin -> verify that you're redirected to the admin page
2. Log in as a teacher -> verify that you're redirected to the teacher page

* Unsure if we should be removing `Default.tsx` entirely - maybe check if user is authenticated and then redirect to login/teacher page/admin page/etc. based on that instead?

## Checklist


frontend/src/App.tsx
```diff
@@ -88,8 +85,6 @@ const App = (): React.ReactElement => {
                   path={Routes.COMPONENT_LIBRARY}
                   component={ComponentLibrary}
                 />
-
-                <Route exact path="*" component={NotFound} />
```
carissa-tang, line 92: is there a reason we are removing this?

frontend/src/App.tsx
```diff
@@ -57,16 +56,14 @@ const App = (): React.ReactElement => {
...
                   exact
-                  path={Routes.HOME_PAGE}
-                  component={Default}
+                  path={["/", Routes.LOGIN_PAGE]}
+                  component={Login}
                 />
+                <Route exact path={Routes.SIGNUP_PAGE} component={Signup} />
 
+                {/* TODO, change to private routes after user management setup */}
                 <PrivateRoute
```
carissa-tang, line 66: might be out of scope for this ticket, but was wondering if teachers have access to `ADMIN_PAGE`? we might need to update the `PrivateRoute` component to validate for specific roles (i.e. role == admin)

frontend/src/components/auth/Login.tsx
```diff
@@ -78,7 +78,11 @@ const Login = (): React.ReactElement => {
   };
 
   if (authenticatedUser) {
-    return <Redirect to={HOME_PAGE} />;
+    return authenticatedUser.role === "Teacher" ? (
```
carissa-tang, line 81: i'm seeing some funky behaviour on my end bc my google email is associated with a role from my previous project that doesn't exist in Jump Math lol (called "donor") and now `authenticatedUser.role == donor` for me. its a very edge case but overall might be better to use a switch statement here specifying only "Teacher" and "Admin" cases + default
```suggestion
        switch (authenticatedUser.role) {
      case "Teacher":
        return <Redirect to={TEACHER_PAGE} />;
      case "Admin":
        return <Redirect to={ADMIN_PAGE} />;
      default:
        return <Redirect to={COMPONENT_LIBRARY} />; // placeholder
    }
```

carissa-tang (commented):
not sure if you saw this message for an unauthenticated user at `\login` + the flow for signing up is broken for me lol
image

also the three buttons (log out, refresh and reset password) aren't working for me but idk if its bc of my weird account

joyce-shi:
> not sure if you saw this message for an unauthenticated user at `\login` + the flow for signing up is broken for me lol image
> 
> also the three buttons (log out, refresh and reset password) aren't working for me but idk if its bc of my weird account

yeah! this is an ongoing issue, if we could fix it with this PR, that would be awesome.

frontend/src/App.tsx
joyce-shi: teachers do not have access to `ADMIN_PAGE`!

teachers should only have access to `TEACHER_PAGE` and admins should only have access to `ADMIN_PAGE`. i think the scope of this ticket is precisely to update the `PrivateRoute` component to validate for specific roles!

frontend/src/components/auth/Login.tsx
joyce-shi: wait LOL, how were you able to sign up as a donor? did you use Google auth? 🧐

carissa-tang: yea my google account - it just auto authenticated me lol (i'm using chrome)

joyce-shi: okay - let's leave this as is for now. I'll look into Google Auth behaviour and create any subsequent tickets to resolve this error! cc @j985chen

frontend/src/App.tsx
chenjanet: nope, whoops lol

chenjanet:
> > not sure if you saw this message for an unauthenticated user at `\login` + the flow for signing up is broken for me lol image
> > also the three buttons (log out, refresh and reset password) aren't working for me but idk if its bc of my weird account
> 
> yeah! this is an ongoing issue, if we could fix it with this PR, that would be awesome.

Made a bug ticket for this because I can't seem to reproduce it locally

frontend/src/components/auth/PrivateRoute.tsx
```diff
@@ -11,17 +12,22 @@ type PrivateRouteProps = {
...
-  ) : (
-    <Redirect to={LOGIN_PAGE} />
-  );
+  if (!authenticatedUser) {
+    return <Redirect to={LOGIN_PAGE} />;
+  }
+  if (
+    (path === TEACHER_PAGE && authenticatedUser.role === "Teacher") ||
+    (path === ADMIN_PAGE && authenticatedUser.role === "Admin")
+  ) {
```
joyce-shi: I'm a bit confused about what's going on here! Does this just hardcode access for specific roles for specific pages? i.e does someone signed in as "Admin" still have access to the `ComponentLibrary`?

joyce-shi (changes requested):
I think we may want to include a `roles` prop that describes which users are allowed to access what? So that in the future, for specific pages, we can describe that there instead of hardcoding access for specific pages? Let me know your thoughts!

frontend/src/components/auth/PrivateRoute.tsx
chenjanet: Yep now that I think about it, your `roles` prop idea does seem like a much better solution - not sure if the code I have right now actually solves the problem of authenticated users only having access to the routes they /should/ based on their roles

frontend/src/App.tsx
```diff
@@ -57,38 +56,37 @@ const App = (): React.ReactElement => {
...
+                  path={[Routes.HOME_PAGE, Routes.LOGIN_PAGE]}
+                  component={Login}
                 />
+                <Route exact path={Routes.SIGNUP_PAGE} component={Signup} />
 
                 <PrivateRoute
                   exact
                   path={Routes.STUDENT_PAGE}
                   component={StudentPage}
+                  roles={["Admin"]}
```
chenjanet, line 70: Ticket mentioned that the student page should be moved to the Admin page; not completely sure whether `Teacher`s should also have access

joyce-shi (approved):
very nice. LGTM! 🚀

PR #60, opened 2022-10-16
feat: remove admin mutation

carissa-tang:
## Notion ticket link

[Remove Admin Mutation](https://www.notion.so/uwblueprintexecs/Remove-Admin-Mutation-48e409f018094baa81921027bebfe861)

## Implementation description
* Add mutation to remove user - calling `deleteUserByEmail`
* Add validation to function - check if there are any GraphQL errors
* Add error message modal
* Updated UI of modal to match Figma

## Steps to test
Checked on Firebase that `c234tang@uwaterloo.ca` (existing user) is removed from the Users table and no longer exists
https://user-images.githubusercontent.com/69697180/196850313-6777d5dd-528a-4e0a-bf4b-36a48b2c984d.mov

When an error occurs while trying to remove a user
https://user-images.githubusercontent.com/69697180/196850644-6cefe154-a448-4a2d-ab90-70a296cf69b0.mov

* Refer to Figma [file](https://www.figma.com/file/cRjJxm8087kkymfH4ieKau/Platform-Working-File?node-id=430%3A12603) for design

## Checklist


frontend/src/components/common/Admin/RemoveUserModal.tsx
```diff
@@ -1,33 +1,76 @@
...
+    variables: { email },
+    fetchPolicy: "cache-and-network",
+    onCompleted: (data: UserGQLObject) => setUser(data.userByEmail.id),
+  });
 
-  const onRemoveUserClick = () => {
-    console.log("Remove user with email: ", email);
-    setShowConfirmation(true);
+  const onRemoveUserClick = async () => {
+    if (user && authenticatedUser?.role === Role.ADMIN) {
```
joyce-shi: do we need the user role check? i don't think non-admins have access to this page

frontend/src/types/AuthTypes.ts
```diff
@@ -11,3 +11,8 @@ export type DecodedJWT =
   | string
   | null
   | { [key: string]: unknown; exp: number };
+
+export enum Role {
```
joyce-shi: why are we changing this from a `type` to an `enum`?

frontend/src/components/common/Admin/RemoveUserModal.tsx
```diff
@@ -1,33 +1,76 @@
...
+
 const RemoveUserModal = ({
   name,
   email,
   onCloseParent,
 }: RemoveUserModalProps): React.ReactElement => {
   const { onOpen, onClose, isOpen } = useDisclosure();
   const [showConfirmation, setShowConfirmation] = React.useState(false);
+  const [showError, setShowError] = React.useState(false);
+  const [removeUser, { error }] = useMutation<{ removeUser: null }>(
```
joyce-shi, line 48: do we need to include `refetchQuery` here? to refresh the list of admin users shown. more context [here](https://www.apollographql.com/docs/react/data/refetching/).

frontend/src/components/common/Admin/RemoveUserModal.tsx
```diff
@@ -1,33 +1,76 @@
...
+const GET_USER = gql`
+  query RemoveUserModal_UserByEmail($email: String!) {
+    userByEmail(email: $email) {
+      id
+    }
+  }
+`;
+
+const REMOVE_USER = gql`
+  mutation RemoveUserModal_DeleteUserByEmail($email: String!) {
```
joyce-shi: maybe we can just call this `DeleteUser` for now. I think we're going to refactor this to the `APIClients` folder down the line for general use, so we can leave it more generic

frontend/src/components/common/Admin/RemoveUserModal.tsx
joyce-shi: similar question for `user`, do we need to refetch and check existence?

joyce-shi (changes requested):

frontend/src/types/AuthTypes.ts
carissa-tang: to use `Role` as a value (i.e. check if role == Role.ADMIN) - but i will remove this since the check is not necessary

frontend/src/components/common/Admin/RemoveUserModal.tsx
carissa-tang: should i make that refactor in this PR?

frontend/src/components/common/Admin/RemoveUserModal.tsx
carissa-tang: hmm i guess validating if the user exists is not needed given that we are working with a table of emails we populated

frontend/src/components/common/Admin/RemoveUserModal.tsx
joyce-shi: no, I think we can make it a separate change, to keep this one modular

frontend/src/components/common/Admin/RemoveUserModal.tsx
joyce-shi: yeah, plus i think if the user doesn't exist, the BE should throw an error anyway

joyce-shi (approved):
🛳️

PR #68, opened 2022-11-10
Admin Users Query 

Boyazhang1:
## Notion ticket link

[Admin Users Query] https://www.notion.so/uwblueprintexecs/47bf4727a2444bccad4a91952898b8f3?v=9658436bf9be409e90581631a7155f16&p=5b06976a90ca44a4b4393c3e59bcf0b2&pm=s

error state
Screen Shot 2022-11-16 at 11 27 36 PM

loading state
Screen Shot 2022-11-16 at 11 27 11 PM

## Implementation description
* created the query on TeacherPage

## Steps to test
1. Go to TeacherPage.tsx, add the line "console.log(adminUsers)"
2. Navigate to localhost:3000/teachers, view the console, and verify the array is in the format
[{firstName, lastName, email}]

* query is formatted properly, handles everything required

## Checklist


frontend/src/components/pages/TeacherPage.tsx
```diff
@@ -1,10 +1,45 @@
...
+import { gql, useQuery } from "@apollo/client";
 
 import Logout from "../auth/Logout";
 import RefreshCredentials from "../auth/RefreshCredentials";
 import ResetPassword from "../auth/ResetPassword";
+import { AdminUser } from "../../types/UserTypes";
 
 const TeacherPage = (): React.ReactElement => {
+  // starts here
+  const ADMIN_USERS = gql`
```
joyce-shi: Sorry, I should have said something earlier, but I think this query already exists under the `RemoveUserModal` component. Could you refactor that to a common folder (maybe create a new `queries` folder under `APIClients`) and then use the same one?

frontend/src/components/pages/TeacherPage.tsx
```diff
@@ -1,10 +1,45 @@
...
+    query AdminUserTable_UsersByRole($role: String!) {
+      usersByRole(role: $role) {
+        firstName
+        lastName
+        email
+      }
+    }
+  `;
+
+  const [adminUsers, setAdminUsers] = useState<AdminUser[]>();
```
joyce-shi: I'm not too sure if we need to use state in this case. I think we may just be able to stick to the style provided by the [documentation](https://www.apollographql.com/docs/react/data/queries#executing-a-query) on Querying in GraphQL? That is, the `const { loading, error, data } = useQuery(...` format. Then, you can just pass `data` to the `AdminTable` component that you built earlier!

Instead of state, `refetchQueries` can be used to handle changes to the list of admin users. This can be called in any mutation that modifies the data that you're querying for here. This should already be implemented in the `RemoveUser` component! You can read more on this [here](https://www.apollographql.com/blog/apollo-client/caching/when-to-use-refetch-queries/).

P.S: Would you be able to check with design to see if there should be loading and error states here?

joyce-shi (changes requested):
Awesome work! 🔥 Left some quick comments. 

One thing: I know we chatted about this earlier, but could we actually put the contents of this query in a separate page? You can just create a new page and name it whatever. Then, when you call the data, you can also pass it to the `AdminTable` component, so that we can test it! 

I know Victor's ticket overlaps a bit, but we probably shouldn't be putting random things in random pages when we don't know where to put them LOL. Totally my bad, I should have said this when we last chatted!

frontend/src/components/pages/AdminTablePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+            size="xl"
+            thickness="4px"
+            emptyColor="gray.200"
+            speed="0.65s"
+          />
+          <Text textStyle="paragraph">
+            Please wait for the data to load. It will load momentarily.
+          </Text>
+        </VStack>
+      </>
```
joyce-shi: what are your thoughts on refactoring the loading state into a variable above?

frontend/src/components/pages/AdminTablePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+      <>
+        <VStack spacing={6}>
+          <Image src={WarningIcon} />
+          <Text textStyle="paragraph">
+            The data has not loaded properly. Please reload the page or contact
+            Jump Math.
+          </Text>
+        </VStack>
+      </>
```
joyce-shi: same with the error state!

frontend/src/components/pages/AdminTablePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+    );
+
+  return (
+    <Box>
+      <AdminUserTable
+        adminUsers={data?.usersByRole?.map((user: AdminUser) => {
+          return {
+            email: user.email,
+            firstName: user.firstName,
+            lastName: user.lastName,
```
joyce-shi: nit: is it necessary to split up `user`? does `data?.usersbyRole?.map((user: AdminUser) => user)` or something similar work?

frontend/src/components/pages/AdminTablePage.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { useQuery } from "@apollo/client";
+import { Spinner, Text, VStack, Image, Box } from "@chakra-ui/react";
+import AdminUserTable from "../common/AdminUserTable";
+import { AdminUser } from "../../types/UserTypes";
+
+import WarningIcon from "../../assets/warning-icon.png";
```
joyce-shi: I think we use the `eva` icon pack in our codebase and design system! Would you be able to import the correct icon from that library into the `icons.tsx` file and use that here?

joyce-shi (changes requested):
left some teeny tiny comments! should only take a min to address (: awesome work!!!

frontend/src/components/pages/AdminTablePage.tsx
Boyazhang1: I think so because the query attaches an id to each user for some reason, but the frontend type doesn't have ID

joyce-shi (approved):
LGTM!

PR #97, opened 2022-12-28
fix: remove hidden field

carissa-tang:
## Notion ticket link

N/A

## Implementation description
* Remove hidden field
* Was initially implemented to update both the school id and school name values based on a single input field, but we don't actually need to store the school name (this is the case that the school already exists)

## Steps to test
What form values look like now - school name will be empty
image

*

## Checklist


joyce-shi (approved):
Yay TY! LGTM

PR #93, opened 2022-12-24
feat: update UserService with new fields

carissa-tang:
## Notion ticket link

[Update userService to store grade selection and currently using fields](https://www.notion.so/uwblueprintexecs/Update-userService-to-store-grade-selection-and-currently-using-fields-db2d028ad4c149feb34d6eb103ca1348)

## Implementation description
* Updated `User` Model to store new fields, UserService implementation, GraphQL types, and tests
* Updated `registerTeacher` resolver to pass new fields to `createUser`

## Steps to test
1. Run yarn `test`
2. Test User service API calls via GraphQL Playground
3. Make test users and check records in Mongo DB
image

when both fields are `undefined` (default)
image

## What should reviewers focus on?
* 

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have run the appropriate linter(s)
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #100, opened 2022-12-28
feat: update register for teacher specific logic + handle submission

carissa-tang:
## Notion ticket link

[Update Register Resolver and Mutation to Handle Teacher Role](https://www.notion.so/uwblueprintexecs/Update-Register-Resolver-and-Mutation-to-Handle-Teacher-Role-74dc25dda05f43f9aec61fa3cdf86bd6)
[Teacher Signup - Page 5 - Submission & Confirmation](https://www.notion.so/uwblueprintexecs/Teacher-Signup-Page-5-Submission-Confirmation-1fe1eeda72034718b53e960436f4646e)

## Implementation description
Register
- Updated `register` auth resolver to take in `role` as a variable
- Renamed REGISTER mutation to REGISTER_TEACHER and pass in “Teacher” as the role
    - Note: I omitted REGISTER_ADMIN for now since there is no use case

Schools
- Added `school` query to get a school by id
- Added GET_SCHOOL query, and CREATE_SCHOOL and ADD_TEACHER_TO_SCHOOL mutations

Frontend
- Removed starter code `Signup` page
- Refactored Steps to access `useForm` methods via FormProvider / useFormContext rather than props
- Page 4 calls `handleSubmitCallback` (function in parent component)  to submit form
- If validation is successful 
    1. Register the new teacher user
    2. Add new teacher to school or create new school
    3. Set authenticated user
- Confirmation page 5

## Steps to test
1. Create new user
image

User in Firebase:
image

User in MongoDB:
image

2a. Add teacher id to selected school (`University of Waterloo`)

MongoDB record - id matches the user's id above
image

2b. Add new school 
image

MongoDB record
image

Confirmation page 5
image

- `firstName`, `lastName`, `grades`, and `currentlyTeachingJM` are not being stored -> will be addressed in https://github.com/uwblueprint/jump-math/pull/93

## Checklist


frontend/src/components/auth/TeacherSignup/TeacherSignup.tsx
```diff
@@ -46,47 +56,117 @@ const renderPageComponent = (
...
+      data.firstName,
+      data.lastName,
+      data.email,
+      data.password,
+      registerTeacher,
+    );
+    if (newUser) {
+      setUser(newUser);
+
+      if (data.school.id) {
```
carissa-tang: quick explanation on why i added the submission logic in this ticket

originally, the ticket was to rewrite `REGISTER_TEACHER` to handle all this logic , but i think due to its complexity, it's easier to handle them as separate calls.

an alternative is to create two mutations 
1. `REGISTER_TEACHER_NEW_SCHOOL` which combines `register` and `createSchool` 
2. `REGISTER_TEACHER_ADD_TEACHER_TO_SCHOOL` which combines `register`, `getSchool`, and `addTeacherToSchool` but I don't think there's a way to combine queries and mutations

current code works, but lmk if you see a cleaner way to implement!

frontend/src/components/auth/TeacherSignup/steps/TeacherSignupFive.tsx
```diff
@@ -0,0 +1,18 @@
...
+const TeacherSignupFive = (): React.ReactElement => {
+  return (
+    <VStack pt={28}>
+      <Text textStyle="header4" textAlign="center" pb={4}>
+        Teacher Sign Up Confirmation
+      </Text>
+      <Text textStyle="subtitle2" textAlign="center">
+        Please check your inbox for a confirmation email so that you are able to
+        login
+      </Text>
```
carissa-tang, line 13: i omitted the `Open Email` button for now bc i'm not sure if there's a direct way to implement this

backend/graphql/resolvers/schoolResolvers.ts
```diff
@@ -12,6 +12,12 @@ const schoolService: ISchoolService = new SchoolService(userService);
 
 const schoolResolvers = {
   Query: {
+    school: async (
+      _req: undefined,
```
joyce-shi: what does `_req` do?

backend/graphql/resolvers/authResolvers.ts
```diff
@@ -33,10 +33,10 @@ const authResolvers = {
     },
     register: async (
       _parent: undefined,
-      { user }: { user: RegisterUserDTO },
+      { user, role }: { user: RegisterUserDTO; role: Role },
       { res }: { res: Response },
     ): Promise<Omit<AuthDTO, "refreshToken">> => {
-      await userService.createUser({ ...user, role: "Teacher" });
+      await userService.createUser({ ...user, role });
```
joyce-shi: isn't `role` a field on `user`? i.e `user.role`? or is there a reason to pass as a separate parameter.

frontend/src/APIClients/mutations/AuthMutations.ts
```diff
@@ -1,7 +1,7 @@
 import { gql } from "@apollo/client";
 
-export const REGISTER = gql`
```
joyce-shi, line 3: hmmm, should we change `ADD_USER` -> `REGISTER_ADMIN` and add it here in `AuthMutations`? I feel like there is a symmetry between those two mutations. Also, should we use the `register` resolver in that mutation as well? 🤔

backend/graphql/types/authType.ts
```diff
@@ -19,7 +19,7 @@ const authType = gql`
 
   extend type Mutation {
     login(email: String!, password: String!): AuthDTO!
-    register(user: RegisterUserDTO!): AuthDTO!
+    register(user: RegisterUserDTO!, role: String!): AuthDTO!
```
joyce-shi: same as above.

frontend/src/components/auth/TeacherSignup/TeacherSignup.tsx
```diff
@@ -42,51 +53,125 @@ const renderPageComponent = (
...
+  const [addTeacherToSchool] = useMutation<{
+    addTeacherToSchool: SchoolResponse;
+  }>(ADD_TEACHER_TO_SCHOOL, {
+    onCompleted() {
+      onSuccessfulSignup();
+    },
+  });
+  const [getSchool] = useLazyQuery<{ school: SchoolResponse }>(GET_SCHOOL, {
+    onCompleted: async (data) => {
+      await addTeacherToSchool({
```
joyce-shi: i'm a bit confused by this! every time we query for a school, do we also call `addTeacherToSchool`?

frontend/src/components/auth/TeacherSignup/TeacherSignup.tsx
joyce-shi: hmm, so i agree that there should be separate calls, but i think that these calls should actually be handled in the backend. 

in my mind, it's similar to when you register a user, the backend will call `createUser` and then trigger `sendVerificationEmail` subsequently. i think the logic to create the associated school, and associate the teacher with that school is akin to the `sendVerificationEmail` part and that `registerTeacher` should trigger those calls! i think this is also better for error handling.

that is, i think here on the frontend, there should only be one call defined and that call should send all the form data to the backend. the backend makes all the calls and then if one part fails, it can avoid being "half-done"

let me know if that makes sense, or if i'm missing something here!

joyce-shi (commented):
left some thoughts!

frontend/src/components/auth/TeacherSignup/steps/TeacherSignupFive.tsx
joyce-shi: yeah! i think that makes sense. we can probably just omit it altogether, pending confirmation from product.

backend/graphql/resolvers/schoolResolvers.ts
```diff
@@ -24,18 +30,6 @@ const schoolResolvers = {
       const newSchool = await schoolService.createSchool(school);
       return newSchool;
     },
-    addTeacherToSchool: async (
```
carissa-tang, line 27: i'm wondering if there's still a separate use case for `createSchool` and `addTeacherToSchool` if i directly implement these in `registerTeacher`? @joyce-shi 

otherwise we can delete?

frontend/src/APIClients/mutations/AuthMutations.ts
carissa-tang: `register` resolver will set the cookies for the current user, but i don't think that would make sense for adding admin since the user != new admin and we only want to create user + send email, which is what `createUser` does

also using the same resolver might not work since the function param `CreateTeacherDTO` would have extra fields unnecessary for `CreateAdminDTO`

but i agree that having them in separate resolvers is a little confusing - would it help to have a `registerTeacher` and `registerAdmin` resolver?

backend/graphql/resolvers/authResolvers.ts
carissa-tang: user of type `RegisterUserDTO` didn't have the role for some reason lol idk why

backend/graphql/resolvers/schoolResolvers.ts
carissa-tang: first arg of a resolver is always the "parent" obj which allows you to return nested objects (via resolver chain) - i believe it can be named anything, but usually smth like `parent`, `obj`, or `_` by convention

https://www.apollographql.com/docs/apollo-server/data/resolvers/#resolver-arguments

joyce-shi: oh nice. thank you!

backend/graphql/resolvers/schoolResolvers.ts
joyce-shi: LOL wait, I feel like it's so sad to delete. But I can't think of any separate use case? I guess if we need them we can always come back to the commit history and find them again D:

frontend/src/APIClients/mutations/AuthMutations.ts
joyce-shi: Ohhhh, I see. For your last point, I think the separate resolvers makes sense to me! To your discretion though, if you think it's fine as is, nbd, I'm excited for your PR to get merged :D

joyce-shi (approved):
LGTM! 🔥

PR #103, opened 2023-01-01
fix: sort popover

joyce-shi:
## Notion ticket link

[Sort Popover Fixes](https://www.notion.so/uwblueprintexecs/Sort-Popover-Fixes-cd19dde3f1724f2b902b257480f786ea)

## Implementation description
* some minor cosmetic fixes to the sort popover

**Updated Screen**
Screen Shot 2022-12-31 at 8 31 53 PM

**Previous Screen**
Screen Shot 2022-12-31 at 8 39 46 PM

**Design Screen** (this is a screenshot of the `Display Assessments` view, design agreed that the admin dashboard view should match this format)
All

### DIFF
--- a/frontend/src/components/common/SortTablePopover.tsx
+++ b/frontend/src/components/common/SortTablePopover.tsx
@@ -44,9 +44,9 @@ const SortTablePopover = ({
           <>
             <PopoverTrigger>
               <Button
+                minWidth="10%"
                 rightIcon={<FilterOptionsIcon />}
                 variant="tertiary"
-                minWidth="10%"
               >
                 Sort
               </Button>
@@ -56,6 +56,7 @@ const SortTablePopover = ({
                 <Box w="1200px">
                   <Stack direction="row" spacing={30} sx={{ pt: "2", pl: "5" }}>
                     <RadioGroup
+                      color="blue.300"
                       onChange={(e) => {
                         if (e === "firstName" || e === "email") {
                           setSortProperty(e);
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/SortTablePopover.tsx
```diff
@@ -44,11 +44,9 @@ const SortTablePopover = ({
           <>
             <PopoverTrigger>
               <Button
-                size="sm"
-                pl="10%"
-                ml="0"
-                rightIcon={<FilterOptionsIcon />}
-                bg="#blue.300"
+                minWidth="10%"
+                leftIcon={<FilterOptionsIcon />}
```
carissa-tang, line 48: not a big deal, but was wondering if we want to use the icon that matches the one in Figma? i can't find it in the eva icons package tho

carissa-tang (commented):
i noticed that the sorting applies directly after you select the option - do we want that behaviour or do we want to use the `Apply Button` and only apply changes when it's clicked?

frontend/src/components/common/SortTablePopover.tsx
joyce-shi: as per design team, we don’t have the icon in the design, so this is the one they wanted to use instead!

joyce-shi:
> i noticed that the sorting applies directly after you select the option - do we want that behaviour or do we want to use the `Apply Button` and only apply changes when it's clicked?

Yup definitely! Good catch. I think the plan here is to refactor this component into a general component (for all tables, not just admin), but to do that I need to do the same for the table component first. So since I’ll be updating the component structure, I can make this change then!

I just wanted to get this merged first so I didn’t forget about some of the smaller cosmetic things I noticed when I was testing the sort feature.

carissa-tang (approved):

PR #111, opened 2023-01-12
feat: table component

joyce-shi:
## Notion ticket link

[Table Component](https://www.notion.so/uwblueprintexecs/Table-Component-b2c0cda0d45c467eb26d106eabebbc47)

## Implementation description
* Refactoring static `AdminUserTable` component to general `Table` component so that we can use it for future tables

We may be able to actually get rid of the `AdminUserTable` component, but I think that might be done later in full refactoring of the `AdminPage`...

### DIFF
--- a/frontend/src/components/common/Table.tsx
+++ b/frontend/src/components/common/Table.tsx
@@ -11,7 +11,7 @@ import {
 
 export interface TableRow {
   values: string[];
-  removeButton: React.ReactElement;
+  menu: React.ReactElement;
 }
 
 interface TableProps {
@@ -50,7 +50,7 @@ export const Table = ({
                   {value}
                 </Td>
               ))}
-              <Td width="5%">{row.removeButton}</Td>
+              <Td width="5%">{row.menu}</Td>
             </Tr>
           ))}
         </Tbody>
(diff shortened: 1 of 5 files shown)


frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,62 @@
...
+  tableRows,
+}: TableProps): React.ReactElement => {
+  return (
+    <TableContainer
+      padding="0.5em"
+      border="1px solid"
+      borderColor="#E2E8F0"
+      borderRadius="12px"
+      minWidth="100%"
+    >
```
joyce-shi, line 30: I couldn't figure out how to extract the css here to a theming file... I think it might not be possible because `TableContainer` is essentially just a `div`, but let me know if this is not the case

carissa-tang (approved):
LGTM! Do we plan to refactor `AdminPage.tsx` so that the tabs section + results is its own component? (aka separate Admin and Teacher)

joyce-shi:
> LGTM! Do we plan to refactor `AdminPage.tsx` so that the tabs section + results is its own component? (aka separate Admin and Teacher)

Hmm... I think so. I think we should probably have a wrapper component for all the Admin pages anyway, since they all have the same sidebar, layout, etc. Not sure exactly what it'll look like yet, but I think `AdminPage` should definitely be split up!

PR #114, opened 2023-01-24
feat: admin dashboard

joyce-shi:
## Implementation description
* Creates an `AdminDashboard` component that wraps all pages shown to the Admin user with the appropriate `VStack` and `NavBar`
* Updated page routing

* No breaking changes
* Routing oopsies


frontend/src/components/pages/admin/UsersPage.tsx
```diff
@@ -0,0 +1,167 @@
+import React from "react";
```
joyce-shi, line 1: No major changes were actually made to this component, I just renamed it from `AdminPage` to `UsersPage` and removed the lines that are now included in `AdminDashboard` (which wraps this component)

frontend/src/components/pages/admin/AdminDashboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+import * as Routes from "../../../constants/Routes";
+import UsersPage from "./UsersPage";
+import CreateQuestionPage from "../../assessment-creation/CreateQuestionPage";
+import PrivateRoute from "../../auth/PrivateRoute";
+import NotFound from "../NotFound";
+
+const pages: Page[] = [
+  { title: "Assessments", url: Routes.ASSESSMENTS },
+  { title: "Database", url: Routes.USER_DATABASE },
+];
```
joyce-shi, line 16: We should probably have some constant object with all the pages urls and components and then map them to the navigation bar and routes... but maybe we can do that when we have more admin routes.

frontend/src/App.tsx
```diff
@@ -93,36 +99,6 @@ const App = (): React.ReactElement => {
                   path={Routes.EMAIL_ACTION}
                   component={EmailActionHandler}
                 />
-                <PrivateRoute
-                  exact
-                  path={Routes.STUDENT_PAGE}
```
carissa-tang, line 98: do we still want `STUDENT_LANDING`

carissa-tang (approved):
lgtm

frontend/src/App.tsx
joyce-shi: maybe we can add when we create the associated page? i think `StudentPage` is empty rn. I kept `TEACHER_LANDING` so that when we login there is somewhere to land

PR #116, opened 2023-01-30
feat: search bar and sort menu components

joyce-shi:
## Notion ticket link

[Table Components](https://www.notion.so/uwblueprintexecs/Menu-Component-f82a589ebae04fcd8923fa4234fcfe9e)

## Implementation description
* Created general `SearchBar` component that we can use in future tables
* Created general `SortMenu` component that we can use in future tables
* Some cosmetic changes to the sort menu to match design changes (i.e dropshadow)

* No breaking changes

Note: my changes now have changed the Sort Menu property to be `FirstName` instead of `Name`, this will be fixed in the next PR where i refactor the `UsersPage` more generally to combine the changes I've made in the last few PRs.

### DIFF
--- a/frontend/src/components/common/SortMenu.tsx
+++ b/frontend/src/components/common/SortMenu.tsx
@@ -37,11 +37,11 @@ const SortMenu = ({
       value={sortProperty}
     >
       <VStack alignItems="left" gap={1}>
-        <Text pb="2" textStyle="link">
+        <Text pb={2} textStyle="link">
           Property
         </Text>
         {tableProperties.map((property, index) => (
-          <Radio defaultChecked={index === 1} key={index} value={property}>
+          <Radio key={index} value={property}>
             {property.charAt(0).toUpperCase() + property.slice(1)}
           </Radio>
         ))}
@@ -59,9 +59,7 @@ const SortMenu = ({
         <Text pb={2} textStyle="link">
           Order
         </Text>
-        <Radio defaultChecked value="ascending">
-          Ascending
-        </Radio>
+        <Radio value="ascending">Ascending</Radio>
         <Radio value="descending">Descending</Radio>
       </VStack>
     </RadioGroup>
--- a/frontend/src/components/common/SortMenu.tsx
+++ b/frontend/src/components/common/SortMenu.tsx
@@ -17,13 +17,13 @@ import {
 import { FilterOptionsIcon } from "../../assets/icons";
 
 interface SortMenuProps {
-  tableProperties: string[];
+  properties: string[];
   onSortProperty: React.Dispatch<React.SetStateAction<string>>;
   onSortOrder: React.Dispatch<React.SetStateAction<string>>;
 }
 
 const SortMenu = ({
-  tableProperties,
+  properties,
   onSortProperty,
   onSortOrder,
 }: SortMenuProps): React.ReactElement => {
@@ -40,7 +40,7 @@ const SortMenu = ({
         <Text pb={2} textStyle="link">
           Property
         </Text>
-        {tableProperties.map((property, index) => (
+        {properties.map((property, index) => (
           <Radio key={index} value={property}>
             {property.charAt(0).toUpperCase() + property.slice(1)}
           </Radio>
--- a/frontend/src/components/common/SortMenu.tsx
+++ b/frontend/src/components/common/SortMenu.tsx
@@ -16,19 +16,56 @@ import {
 } from "@chakra-ui/react";
 import { FilterOptionsIcon } from "../../assets/icons";
 
-interface TableSortProps {
-  properties: string[];
+interface SortMenuProps {
+  tableProperties: string[];
   onSortProperty: React.Dispatch<React.SetStateAction<string>>;
   onSortOrder: React.Dispatch<React.SetStateAction<string>>;
 }
 
-const SortTablePopover = ({
-  properties,
+const SortMenu = ({
+  tableProperties,
   onSortProperty,
   onSortOrder,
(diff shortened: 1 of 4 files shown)


frontend/src/components/common/SortMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+              </PopoverFooter>
+            </PopoverContent>
+          </>
+        )}
+      </Popover>
+    </>
+  );
+};
+
+export default SortMenu;
```
joyce-shi, line 118: This is `SortTablePopover` with some changes, but the renaming made it a new file.

frontend/src/components/pages/admin/UsersPage.tsx
```diff
@@ -131,19 +123,12 @@ const UsersPage = (): React.ReactElement => {
...
-                        placeholder="Search bar"
-                      />
-                      <InputRightElement pointerEvents="none" h="full">
-                        <SearchOutlineIcon />
-                      </InputRightElement>
-                    </InputGroup>
-                    <SortTablePopover OrderingSets={OrderingSets} />
+                    <SearchBar onSearch={setSearch} />
+                    <SortMenu
+                      tableProperties={["firstName", "email"]}
```
joyce-shi: In the next PR I think tableProperties will have the same value as headers in the AdminUserTable (and AdminUserTable can be removed as a separate file.

frontend/src/components/common/SortMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+                  </HStack>
+                </Flex>
+              </PopoverBody>
+              <PopoverFooter border="0" alignSelf="end" px="10">
+                <Button
+                  minWidth="10%"
+                  onClick={() => {
+                    onSortProperty(sortProperty);
+                    onSortOrder(sortOrder);
+                    onClose();
```
joyce-shi, line 103: nit: I want to move this into a separate handler function... but I don't know how to define `onClose` outside of the JSX and passing the function as an argument seems like overkill

carissa-tang (commented):
not sure if i'm looking at the right designs, but should the search and sort bar be above the tabs?

joyce-shi:
> not sure if i'm looking at the right designs, but should the search and sort bar be above the tabs?

i think there's several iterations! but according to most recent sync w/ design, search and sort should be below the tabs. (the hierarchy of elements in the users page should match the hierarchy of elements in the `Display Assessments` page

frontend/src/components/pages/admin/UsersPage.tsx
carissa-tang: is there a reason we pass the values like so? rather than `["Name", 
"Email"]  - these will be consistent with both the table headers and sort properties

then we can avoid the workarounds in line 45 of `SortMenu`

frontend/src/components/common/SortMenu.tsx
```diff
@@ -0,0 +1,120 @@
...
+      color="blue.300"
+      onChange={(e) => setSortProperty(e)}
+      value={sortProperty}
+    >
+      <VStack alignItems="left" gap={1}>
+        <Text pb="2" textStyle="link">
+          Property
+        </Text>
+        {tableProperties.map((property, index) => (
+          <Radio defaultChecked={index === 1} key={index} value={property}>
```
carissa-tang: could also do the following to adjust the UI
```suggestion
          <Radio defaultChecked={index === 1} key={index} value={property} style={{ textTransform: "capitalize" }} >
```

frontend/src/components/pages/admin/UsersPage.tsx
joyce-shi: It's because `adminUser` has `firstName`, `lastName` and `email` as fields, so you can have to sort by one of those. I think in the (hopefully) last refactor for the `UsersPage` I would concatenate the first two fields and have the `AdminUser` data stored in an object that just has a Name` and `Email` field. Let me know if that makes sense!

frontend/src/components/common/SortMenu.tsx
joyce-shi: OH beautiful! Will add, thank you!

carissa-tang: should this be 0? but also, i removed this attribute and it was same behaviour - so i'm not sure if it's actually doing anything ?? lol

frontend/src/components/pages/admin/UsersPage.tsx
carissa-tang: ohh makes sense !

frontend/src/components/common/SortMenu.tsx
joyce-shi: good catch! not needed

joyce-shi: Just kidding. I think `Radio` is not technically a text component, so doesn't seem like applying the `textTransform` attribute works here. Options here: (1) leave code as is (2) wrap the children text in a `Text` object and apply the `textTransform`. Thoughts on preference here?

carissa-tang: ohh hmmm we can just leave as is then

though i noticed that "firstName" is now displayed as "FirstName" instead of "First Name" but ig it's fine lol cause splitting that might complicate the code

carissa-tang (approved):

frontend/src/components/common/SortMenu.tsx
joyce-shi: yeah - i think when we change it to `Name` we won't have to worry about that anyway!

PR #139, opened 2023-02-28
feat: question element for question editor

joyce-shi:
## Notion ticket link

[Add question component to Question Editor](https://www.notion.so/uwblueprintexecs/Add-question-component-to-Question-Editor-f28efb8aaf8b478d84899013af9e6b44?pvs=4)

## Implementation description
* added question element to question editor, has auto-lettering
* minor cosmetic fixes in other areas

* code cleanliness
* cosmetic issues


frontend/src/themes/components/button.ts
```diff
@@ -45,6 +45,10 @@ const Button = {
       },
       margin: "0 10px",
     },
+    icon: {
+      minWidth: "100%",
+      padding: "0px",
+    },
```
joyce-shi, line 51: reduces width of icon buttons

frontend/src/components/common/QuestionCard.tsx
```diff
@@ -78,7 +78,7 @@ const QuestionCard = ({
           </Flex>
           <List
             spacing={3}
-            styleType="lower-alpha"
+            styleType={questions.length > 1 ? "lower-alpha" : "none"}
```
joyce-shi, line 81: questions with only one sub-question should not be numbered. (i.e should not have `a.` before the question if there is only one question)

frontend/src/components/question-creation/question-elements/TextElement.tsx
```diff
@@ -33,6 +33,7 @@ const TextElement = ({ id, data }: TextElementProps): React.ReactElement => {
 
   return (
     <Textarea
+      sx={{ fontSize: "18px", fontWeight: "400", lineHeight: "23px" }}
```
joyce-shi: couldn't get `textStyle` to work here... so manually added typography 😟 let me know if you have thoughts on how to use `textStyle` to work instead!

frontend/src/components/question-creation/question-elements/TextElement.tsx
```diff
@@ -43,6 +44,7 @@ const TextElement = ({ id, data }: TextElementProps): React.ReactElement => {
       resize="none"
       minRows={1}
       as={ResizeTextarea}
+      paddingLeft="6"
```
joyce-shi, line 35: this differs from the design, but design asked to have all elements other than the question element be further indented

frontend/src/components/question-creation/question-elements/TextElement.tsx
carissa-tang: hmm maybe you could look into defining variants in `themes/components/textArea.ts`?

frontend/src/components/question-creation/question-elements/QuestionTextElement.tsx
```diff
@@ -0,0 +1,76 @@
...
+  ).length;
+  const questionLetter = String.fromCharCode(
+    "a".charCodeAt(0) +
+      (questionElements
+        .filter((element) => element.type === QuestionElementType.QUESTION)
+        .findIndex((element) => element.id === id) %
+        26),
+  );
+
+  const updateQuestionElement = (updatedQuestion: string) => {
```
carissa-tang: if this is going to be reused for each component, should we also factor it out? and have it accept an error message as input

i'm not sure what the best practice is here, bc if we are defining this function in `QuestionElementItem.tsx` and passing down as props, that defeats the whole context thing? but then having 6 copies of this function also seems wrong

carissa-tang (approved):
it's coming togetherrr so exciting !!

just left some thoughts

frontend/src/components/question-creation/question-elements/QuestionTextElement.tsx
joyce-shi: +1! i'll refactor it out. I agree, I don't think I should be passing context down as props, so I think I might put it in `QuestionUtils`?

joyce-shi: i moved the logic into `QuestionUtils` but left the setter in the individual elements! let me know if that's better

frontend/src/utils/QuestionUtils.ts
```diff
@@ -32,3 +29,20 @@ export const shouldReorder = (
...
+  updatedData: string,
+  error: string,
+  prevElements: QuestionElement[],
+) => {
+  const indexToUpdate = prevElements.findIndex((element) => element.id === id);
+  return update(prevElements, {
+    [indexToUpdate]: {
+      $merge: {
+        data: updatedData,
+        error: updatedData.length > 800 ? error : "",
```
carissa-tang: hmm maybe we move the error logic into the components themselves? since this length check only applies to text and question text, but if we want to use for other components will be different

frontend/src/components/question-creation/question-elements/QuestionTextElement.tsx
carissa-tang: this is good !!

frontend/src/utils/QuestionUtils.ts
joyce-shi: oops I’m silly, ty!

PR #145, opened 2023-03-05
fix: Delete dependent documents on deleteTeacher

carissa-tang:
## Notion ticket link

[Delete Test Sessions associated with teacher on deleteTeacher](https://www.notion.so/uwblueprintexecs/Delete-Test-Sessions-associated-with-teacher-on-deleteTeacher-38743ea27b1648cd9ea36e7df5d6bf31?pvs=4)
[Delete teacher id from associated School Object on deleteTeacher](https://www.notion.so/uwblueprintexecs/Delete-teacher-id-from-associated-School-Object-on-deleteTeacher-f6595a11dd6f46fcba0c250da9f203ab?pvs=4)

## Implementation description
* Updated `UserModel` middleware to do the following after `findOneAndDelete` is executed
   * Delete all test sessions containing the deleted teacher's id
   * Delete the teacher id from the School document which contains it
* Wrote unit tests for success and failure

## Steps to test
1. yarn test

* Info on middleware: https://mongoosejs.com/docs/middleware.html#use-cases

## Checklist

### DIFF
--- a/backend/models/user.model.ts
+++ b/backend/models/user.model.ts
@@ -64,16 +64,19 @@ const UserSchema: Schema = new Schema({
   },
 });
 
-UserSchema.post("findOneAndDelete", async (doc) => {
+UserSchema.pre("findOneAndDelete", async function (next) {
+  const doc = await this.findOne(this.getQuery());
   if (doc.role !== "Teacher") return;
 
   /* eslint-disable no-underscore-dangle */
-  await MgTestSession.deleteMany({ teacher: doc._id });
   await MgSchool.findOneAndUpdate(
     { teachers: doc._id },
     { $pull: { teachers: doc._id } },
     { new: true },
   );
+  await MgTestSession.deleteMany({ teacher: doc._id });
+
+  next();
 });
 
 export default mongoose.model<User>("User", UserSchema);
--- a/backend/models/user.model.ts
+++ b/backend/models/user.model.ts
@@ -1,4 +1,6 @@
 import mongoose, { Schema, Document } from "mongoose";
+import MgTestSession from "./testSession.model";
+import MgSchool from "./school.model";
 
 import { Grade, Role } from "../types";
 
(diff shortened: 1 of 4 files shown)


backend/services/implementations/__tests__/userService.test.ts
```diff
@@ -94,26 +102,16 @@ describe("mongo userService", (): void => {
   it("getAllTeachers", async () => {
     const createRes = await UserModel.insertMany(testUsers);
 
-    const testSchools = [
```
carissa-tang, line 97: refactor to use existing mock data

backend/models/user.model.ts
```diff
@@ -62,4 +64,16 @@ const UserSchema: Schema = new Schema({
...
+  if (doc.role !== "Teacher") return;
+
+  /* eslint-disable no-underscore-dangle */
+  await MgTestSession.deleteMany({ teacher: doc._id });
+  await MgSchool.findOneAndUpdate(
+    { teachers: doc._id },
+    { $pull: { teachers: doc._id } },
+    { new: true },
+  );
+});
```
joyce-shi, line 80: bit of a nit: but should we use `pre` hook here instead of `post`? i.e delete child dependencies first and then the parent?

joyce-shi (approved):
lgtm! left one minor q

backend/models/user.model.ts
carissa-tang: hmm that makes sense - but i was also thinking if user deletion fails, would we need to roll back the changes?

joyce-shi: LOL I was actually also thinking about that when I added this comment, i.e if deletion fails, is it better to have undeleted test sessions or undeleted teacher? 👀

i.e if you deleted the test session and it worked, but then the teacher didn't delete, it would still be fine (b/c teacher would just have one less test session). but if deleted teacher & the test session didn't delete, then there would be a test session w/ an undefined teacher field... and not sure if we handle that as gracefully? but maybe our code should just handle all those cases gracefully? idk

TLDR i'm not actually sure what is best here 😢

carissa-tang: i think that makes sense. so we delete the dependencies before the teacher. i can look into rolling back the changes for that too

PR #153, opened 2023-03-09
feat: add multiple choice element

joyce-shi:
## Notion ticket link

[Add multiple choice component to Question Editor](https://www.notion.so/uwblueprintexecs/Add-multiple-choice-component-to-Question-Editor-47701f7b45014849afd96be6fd365be2?pvs=4)

## Implementation description
* Created a general modal for responses called `ResponseTypeModal`
* Refactored Short Answer modal to use `ResponseTypeModal`
* Implemented Multiple Choice element

## Steps to test
1. Test in staging

* Add and edit Multiple Choice functionality
* Selecting number of options for Multiple Choice functionality (& option deletion)
* Error states for Multiple Choice modal
* Code cleanliness and best practices

Note: some differences from design: 
- design has asked to have response state filled (i.e correct answer is shown as selected)
- error state diffs are pending design approval


carissa-tang (commented):
when i switch between the # of options, it gets out of sync

https://user-images.githubusercontent.com/69697180/224460302-5f9f163a-92c7-4be1-a65c-0e55becf0d8e.mov

also can we add wrap text here?
image

and does it make sense to allow only 1 option? lol

also i believe there's 800char limit?

joyce-shi:
> when i switch between the # of options, it gets out of sync
> 
> also can we add wrap text here? image
> 
> and does it make sense to allow only 1 option? lol
> 
> also i believe there's 800char limit?

oooo good catch! i made some changes and i _think_ i resolved the options bug? but i had some trouble reproducing it, so let me know if that's not the case! 

re: wrap text, i think it already wraps? see image. but let me know if i'm not understanding something here!
Screen Shot 2023-03-11 at 3 21 23 AM

re: 800 character limit, added the 800 character limit to the modal error handling to the modal! it differs from design, but i dont think it makes sense to allow user to add and then show the error...

re: 1 option - i think there might be a use case for it? like if the first question is "confirm that you will not plagiarize" & then theres only one option thats `yes` LOL sort of thing?

frontend/src/components/question-creation/question-elements/modals/multiple-choice/SelectOptionCount.tsx
```diff
@@ -0,0 +1,71 @@
...
+      [...Array(elementsToAdd)].forEach(() =>
+        addOption({ id: uuidv4(), value: "", isCorrect: false }),
+      );
+    }
+    if (count < optionCount) {
+      const elementsToRemove = optionCount - count;
+      [...Array(elementsToRemove)].forEach(() => removeLastOption());
+    }
+    setOptionCount(count);
+  };
```
carissa-tang:
```suggestion
  const handleSelectCount = (event: React.ChangeEvent<HTMLSelectElement>) => {
    const count = parseInt(event.target.value, 10);
    const countDiff = count - optionCount;
    if (countDiff > 0) {
      [...Array(countDiff)].forEach(() =>
        addOption({ id: uuidv4(), value: "", isCorrect: false }),
      );
    } else {
      setOptions((prevOptions) => prevOptions.slice(0, countDiff));
    }
    setOptionCount(count);
  };
```

frontend/src/components/question-creation/question-elements/modals/multiple-choice/MultipleChoiceOption.tsx
```diff
@@ -0,0 +1,105 @@
...
+  data,
+  isEmptyError,
+  isCorrectError,
+  setOptions,
+  setOptionCount,
+}: MultipleChoiceOptionProps): React.ReactElement => {
+  const { id, value, isCorrect } = data;
+  const [lengthError, setLengthError] = useState(false);
+
+  const isOversized = (input: string) => input.length > 800;
```
carissa-tang:
```suggestion
  const exceedsMaxLen = (input: string) => input.length > 800;
```

carissa-tang: if we're extracting this into a function, do we want to put it into utils and refactor the other elements to use it? otherwise i think we can directly embed this logic?

carissa-tang (approved):
just left some suggestions. but lgtm!

frontend/src/components/question-creation/question-elements/modals/multiple-choice/MultipleChoiceOption.tsx
joyce-shi: put into utils & updated other components to use it! i think its a bit more readable this way

PR #160, opened 2023-03-11
chore: folder reorg + some refactoring + minor fixes

joyce-shi:
## Notion ticket link

Trying to bring some order to the chaos...you may want to pull the branch instead of trying to sort through `Files changed` 😅 
[Fix status tag capitalization](https://www.notion.so/uwblueprintexecs/Fix-Status-Tag-capitalization-2bda9b67b5334dba9bb3a4728ecc3d9c?pvs=4)

## Implementation description
* fixed capitalization issue with status tag & type of assessment field in assessment table
* refactored some errors to use common components
* added some folders to organize

* no breaking changes
* let me know if there are any suggestions to make this spring cleaning better LOL


joyce-shi:
cc @wais-hundekar fixed the capitalization issues!

frontend/src/types/AssessmentTypes.ts
```diff
@@ -1,10 +1,20 @@
-export type StatusProperty = "DRAFT" | "PUBLISHED" | "ARCHIVED" | "DELETED";
+export enum Status {
+  DRAFT = "Draft",
```
carissa-tang: in the backend we store the enums capitalized like so, and with different variable names https://github.com/uwblueprint/jump-math/blob/staging/backend/models/test.model.ts#L15. can we make sure they're consistent?

joyce-shi: good catch! updated the enums & then the fe to format properly. also added a general back button component, since i think we'll use it again!

frontend/src/types/AssessmentTypes.ts
```diff
@@ -1,10 +1,20 @@
-export type StatusProperty = "DRAFT" | "PUBLISHED" | "ARCHIVED" | "DELETED";
+export enum Status {
```
carissa-tang, line 1: do we want to update the BE variable names to match? `AssessmentStatus -> Status` for `AssessmentType -> UseCase`

in a follow up pr

frontend/src/components/common/BackButton.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
```
carissa-tang, line 1: can `QuestionSidebar` also use this?

joyce-shi: yes maaaam, but i refactored the `QuestionSidebar` in my other open PR (to get rid of the props + history in the file), so after I merge this one I was gonna make the update there 😅

frontend/src/types/AssessmentTypes.ts
joyce-shi: sounds good, will do! i think it might be a good idea to create a ticket to consolidate our types in general sometime soon 🤔

frontend/src/components/common/BackButton.tsx
carissa-tang: its fine if the refactor is not clean

carissa-tang (approved):

frontend/src/components/common/BackButton.tsx
joyce-shi: adding!

PR #162, opened 2023-03-12
backend: Convert Grade Field to Enum

wais-hundekar:
## Notion ticket link

[Convert Grade Field To Enum](https://www.notion.so/uwblueprintexecs/Convert-Grade-Field-To-Enum-631a1b09650d4fc9a4ccff74f488b02e?pvs=4)

## Implementation description
* Refactoring throughout the codebase so Grade Enum is used
* Changed certain options to use Object.keys(Grade) for readability
* Updated the options for AddClassroomModal to use all Grade options
* Updated testCase function to also remove underscores
Screenshot 2023-03-12 at 1 48 25 AM

Successful Mutation

## Steps to test
1. Ensure all mutations still work
2. Ensure tests are fixed

* It is important that no existing functionality is broken due to this change

## Checklist


backend/models/test.model.ts
```diff
@@ -124,8 +125,19 @@ const TestSchema: Schema = new Schema(
       required: true,
     },
     grade: {
-      type: Number,
+      type: String,
       required: true,
+      enum: [
```
carissa-tang: hmm could you actually make an enum for grade and then refactor all the models to use it (i.e. `enum: Object.keys(Grade),`

wais-hundekar: I could, but then I'd also have to make the objects GRADE_1 for example, I was wondering if that'd be a problem for consistency with the other files that use grade as a string? Or should I be changing those files as well?

wais-hundekar: Like the `user.model.ts` file as well?

carissa-tang: which files are you referring you? if its like "Grade 1", they should convert to `Grade.ONE` or smth

carissa-tang: yes update user.model.ts and class.model.ts

wais-hundekar: Done!

backend/graphql/types/userType.ts
```diff
@@ -6,13 +6,25 @@ const userType = gql`
     Admin
   }
 
+  enum UserGradeEnum {
```
carissa-tang:
```suggestion
  enum GradeEnum {
```
for consistency

backend/graphql/types/testType.ts
```diff
@@ -22,6 +22,18 @@ const testType = gql`
...
+    GRADE_1
+    GRADE_2
+    GRADE_3
+    GRADE_4
+    GRADE_5
+    GRADE_6
+    GRADE_7
+    GRADE_8
+  }
+
```
carissa-tang:
```suggestion
```
we only need to define it in one file! since it'll be all combined in the end so it can read the other file types

carissa-tang (commented):
these are some of the failed test errors (i believe you need to update all our mock data to use the enum). you can do ctrl f to find all places with `"Grade ` (both FE and BE)

image

backend/graphql/types/testType.ts
wais-hundekar: Oh the more you know

frontend/src/constants/CreateAssessmentConstants.ts
```diff
@@ -1,40 +1,8 @@
...
-  {
-    label: "Grade 8",
-    value: "8",
-  },
-];
+import { Grade } from "../APIClients/types/UserClientTypes";
+
+const gradeOptions = Object.keys(Grade).map((grade) => ({
+  value: grade,
+  label: grade,
```
wais-hundekar: I now noticed I forgot to add titlecase to this label, will do so with any additional review changes in the next commit

joyce-shi (changes requested):
in the enum definitions, instead of defining grades as `GRADE_1` or `GRADE_2`, etc., can we just define it as 1 or 2?

i.e instead of `grade: Grade.GRADE_1` we can write `grade: Grade.1`, to keep it a bit cleaner. additionally, instead of adding the removing `_` logic in the shared `titleCase` function, we can just append `Grade` to the front of the table entry, so that we keep the `titleCase` function to single purpose

wais-hundekar:
> in the enum definitions, instead of defining grades as `GRADE_1` or `GRADE_2`, etc., can we just define it as 1 or 2?
> 
> i.e instead of `grade: Grade.GRADE_1` we can write `grade: Grade.1`, to keep it a bit cleaner. additionally, instead of adding the removing `_` logic in the shared `titleCase` function, we can just append `Grade` to the front of the table entry, so that we keep the `titleCase` function to single purpose

@joyce-shi An enum member cannot have a numeric name, should I change it to something like Grade._1?

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -153,16 +155,10 @@ const AddClassroomModal = (): React.ReactElement => {
...
-                      {
-                        value: "grade 1",
-                        label: "grade 1",
-                      },
-                      {
-                        value: "grade 2",
-                        label: "grade 2",
-                      },
-                    ]}
+                    options={Object.keys(Grade).map((grade) => ({
```
carissa-tang: can we extract this into a constant that can be used across the files (e.g. `frontend/src/constants/CreateAssessmentConstants.ts`)

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
```diff
@@ -137,19 +139,9 @@ const TeacherSignupOne = ({
...
-              "Grade 3",
-              "Grade 6",
-              "Grade 1",
-              "Grade 4",
-              "Grade 7",
-              "Grade 2",
-              "Grade 5",
-              "Grade 8",
-            ].map((grade) => (
+            {Object.keys(Grade).map((grade) => (
```
carissa-tang: i think you can also use the constant i mentioned below to map values to Checkbox

joyce-shi:
> 

In that case, I think you can set the value to the number, i.e `GRADE_1 = 1`.

joyce-shi:
> > in the enum definitions, instead of defining grades as `GRADE_1` or `GRADE_2`, etc., can we just define it as 1 or 2?
> > i.e instead of `grade: Grade.GRADE_1` we can write `grade: Grade.1`, to keep it a bit cleaner. additionally, instead of adding the removing `_` logic in the shared `titleCase` function, we can just append `Grade` to the front of the table entry, so that we keep the `titleCase` function to single purpose
> 
> @joyce-shi An enum member cannot have a numeric name, should I change it to something like Grade._1?

In that case, I think you can set the value to the number, i.e `GRADE_1 = "1"`.

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
wais-hundekar: Done!

frontend/src/components/user-management/student/AddClassroomModal.tsx
wais-hundekar: Done!

carissa-tang:
> > > in the enum definitions, instead of defining grades as `GRADE_1` or `GRADE_2`, etc., can we just define it as 1 or 2?
> > > i.e instead of `grade: Grade.GRADE_1` we can write `grade: Grade.1`, to keep it a bit cleaner. additionally, instead of adding the removing `_` logic in the shared `titleCase` function, we can just append `Grade` to the front of the table entry, so that we keep the `titleCase` function to single purpose
> > 
> > 
> > @joyce-shi An enum member cannot have a numeric name, should I change it to something like Grade._1?
> 
> In that case, I think you can set the value to the number, i.e `GRADE_1 = "1"`.

@joyce-shi @wais-hundekar if we want to change the naming, we will need to add this to the test resolver
```
  GradeEnum: {
    K: "K",
    GRADE_1: "1",
    GRADE_2: "2",
    GRADE_3: "3",
    GRADE_4: "4",
    GRADE_5: "5",
    GRADE_6: "6",
    GRADE_7: "7",
    GRADE_8: "8",
  },
    Query: {
    ...
  ```

joyce-shi (approved):
LGTM, good stuff

PR #166, opened 2023-03-16
query: Student Login Query 

JennaXiao3:
## Notion ticket link

Create query for Student Authentication (https://www.notion.so/uwblueprintexecs/Create-query-for-Student-Authentication-60f4de968b3541748668e231975649a9?pvs=4)

## Implementation description
* created TestSessionQueries.ts with query to get test session by access code
* added necessary code to Student Login to ensure that border colour changes depending on code success/failure

## Steps to test
1. Go to http://localhost:3000/student-login
2. Check if border colour is red with invalid access code (ex. 123329)
3. Check if border colour is green with valid access code (ex. 888888)

* right now, GET_TEST_SESSION_BY_ACCESS_CODE returns just id and accessCode since right now I don't need any info (since I am just checking if access code is valid) but this might change in the future - should I change it?
* when you enter the wrong code, border colour is red and error message is displayed --> if you click on any pin after this, the entire pin is reset --> is this okay?

## Checklist


frontend/src/components/auth/student-login/StudentLogin.tsx
```diff
@@ -12,33 +20,72 @@ const StudentLogin = (): React.ReactElement => {
...
         >
           {[...Array(6)].map((i) => (
             <PinInputField
               key={i}
               _focus={{ backgroundColor: "grey.100" }}
               _hover={{ backgroundColor: "grey.100" }}
-              _invalid={{ borderColor: "red.200" }}
-              _valid={{ backgroundColor: "green.200" }}
               backgroundColor="grey.100"
+              borderColor={pinBorderColor}
```
joyce-shi: nit: I think instead of using `pinBorderColor` as a separate state, since it entirely depends on `success` and `error` values, you can just create a function above which queries for the value of `borderColor`! then can avoid the use of the `!important` override

i.e define a `getPinBorderColor()`

frontend/src/components/auth/student-login/StudentLogin.tsx
```diff
@@ -12,33 +20,72 @@ const StudentLogin = (): React.ReactElement => {
...
               color="grey.300"
               fontSize="2.5rem"
               height="50%"
               m="3rem 0.6rem 7rem 0.6rem !important"
+              onFocus={resetPin}
               style={{ width: "15%" }}
               textStyle="header1"
+              value={i}
```
joyce-shi: what does this do?

frontend/src/APIClients/queries/TestSessionQueries.ts
```diff
@@ -0,0 +1,12 @@
+import { gql } from "@apollo/client";
+
+const GET_TEST_SESSION_BY_ACCESS_CODE = gql`
+  query TestSessionByAccessCode($accessCode: String!) {
+    testSessionByAccessCode(accessCode: $accessCode) {
+      id
+      accessCode
```
joyce-shi: I don't think we need to return `accessCode`!

joyce-shi (commented):
left some small comments! overall looks amazing! 🎉 can't wait to get this merged :3

frontend/src/components/auth/student-login/StudentLogin.tsx
JennaXiao3: oh nice catch! i forgot to remove it

frontend/src/components/auth/student-login/StudentLogin.tsx
JennaXiao3: whwere would getPinBorderColor() get called from? the reason why !important is there is to override the style for the focus state (which is a gray.200). Once the user finishes the input, the focus is still on the last pin input, so without !important, only the last cell will be gray bordered while the rest is the right colour

joyce-shi: oooo okay makes sense! i think ideally, after the pin completes, none of the fields should be focused? then shouldn't need to override! but i think if this is difficult, we can just use the `!important` in the function.

re: where to call the function - you can set `borderColor={getPinBorderColor()}` as a `PinInputField` attribute~! & then it will dynamically assign `borderColor`. i think even with the `!important` attribute, we want to try keep data stored in states, not the attributes derived from that data! lmk if you have thoughts

joyce-shi: also, small bug i just noticed is that i think there might be some issue w/ the focusing? 🤔 i noticed that it doesnt input numbers where my cursor is, i.e if i set the cursor to the last pin input, it still fills the first non-empty one. I think this is because the pin fields should be setting input elements individually! i.e the third `PinInputField` should update the value of pin[3], not the whole string.

https://user-images.githubusercontent.com/41273486/225983102-bb3f2d9b-a8e6-43b6-a73d-0bf27f106c08.mov

JennaXiao3: i was able to remove the focus but it only happens like 10 seconds after completion of input, so I have removed it and kept the "!important" if that is okay. everything else should be fine!

joyce-shi: i'm still seeing the bug described above? let me know if you're having trouble reproducing it!

JennaXiao3: Oh sorry I misunderstood ur comment. I think that's a problem with mui's pin input, since there is a wrapper for all the pininputfields called pininput which takes in the whole entire value. But to avoid this problem, I can try to disable all the other inputs before the user inputs? would that work?

Edit: actually it might've been something i did...i will look into it more tomorrow i apologize for this

joyce-shi: sounds good! no worries at all. we can work through it together if youd like, just lmk! (:

JennaXiao3: should be fixed now!

joyce-shi (approved):
LGTM!! awesome work 🛳

PR #173, opened 2023-03-23
feat: save question editor data

joyce-shi:
## Notion ticket link

[Save Question Editor data to Create Assessment page](https://www.notion.so/uwblueprintexecs/Save-Question-Editor-data-to-Create-Assessment-page-c95b204965a341499bf3657f903ebd68?pvs=4)

## Implementation description
* refactored `QuestionPage` from standalone page to be a child page of `CreateAssessmentPage` (it should never be accessed directly without a parent `CreateAssessmentPage` anyway)
* added date as dynamic route to `CreateAssessmentPage` (this prevents the user from directly accessing this page and fixes the earlier date bug)
* removes the `three-dot popover` from the `CreateAssessmentHeader` since it will be unused
* saves `QuestionEditor` data to `CreateAssessmentPage`
* dynamically renders `QuestionCards` according to saved data
* refactored `QuestionType` -> `ResponseElementType`, as a subset of `QuestionElementType` (to link the work that was done on the `QuestionCard` component to the `QuestionEditor` work)
* added error validation to the question editor for the following cases:
    * no empty editor
    * no empty element (i.e can't leave text blank in `QuestionTextElement`, etc.)
    * no improper question pairings (i.e all questions must have responses, all responses have questions)
    * no existing errors (i.e text length errors)

## Steps to test
1. test on staging env

* feature is implemented in accordance with design
* no breaking changes
* check that error validation works properly (i'm not 100% confident in this)


frontend/src/types/QuestionTypes.ts
```diff
@@ -13,6 +7,29 @@ export enum QuestionElementType {
...
 
+export type ResponseElementType = Extract<
+  QuestionElementType,
+  | QuestionElementType.MULTIPLE_CHOICE
+  | QuestionElementType.SHORT_ANSWER
+  | QuestionElementType.MULTI_SELECT
+>;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export const ResponseElementType = {
```
carissa-tang, line 18: is there a reason we aren't using a `ResponseElementEnum`? might be cleaner to use like `ResponseElementEnum. MULTIPLE_CHOICE ` for example in `QuestionTag.tsx` since type is restricted to `ResponseElementType`

frontend/src/components/question-creation/question-elements/SaveQuestionEditorButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+        errorText,
+      );
+    });
+  };
+
+  const isQuestionPairError = (elementError: string) => {
+    return elementError === questionError || elementError === responseError;
+  };
+
+  const validateNoQuestionPairErrors = () => {
```
carissa-tang, line 49: does it have to be a question specifically? or can it also be text / image

carissa-tang, line 49: also can you just have text / image alone without a response? bc then it creates its own question card that's blank
image

frontend/src/components/assessments/assessment-creation/CreateAssessementButton.tsx
```diff
@@ -1,33 +1,25 @@
...
   return (
-    <>
-      <Button
-        my={2}
-        onClick={navigateToCreateAssessmentPage}
-        rightIcon={<PlusOutlineIcon />}
-        variant="primary"
-      >
+    <Button my={2} rightIcon={<PlusOutlineIcon />} variant="primary">
+      <RouterLink to={`${CREATE_ASSESSMENT}/${formattedDate}`}>
```
carissa-tang: is there a reason we use the formatted date here?

design-wise it looks a little funky to me

also idk if its an issue if ppl can tamper the url (e.g. manually set to `March%2026,%202022`) - maybe there's a better way to hash the value? just some thoughts
image

frontend/src/types/QuestionTypes.ts
joyce-shi: i.e why don't we use a separate unrelated enum entirely?

here, i've defined `ResponseElementType` as a subset of `QuestionElementType`. i think this is beneficial in cases like `generateQuestionCardTags()` where it can take elements from the question editor (`QuestionElementType`) and map it to a specific `QuestionTag` (of `ResponseElementType).

let me know if that makes sense or if i've misunderstood your comment!

frontend/src/components/question-creation/question-elements/SaveQuestionEditorButton.tsx
joyce-shi: good catch! i'll add handling for this

frontend/src/types/QuestionTypes.ts
carissa-tang: ah ok - i guess just in general i was thinking if we should have two enums for "Prompt" types (Question, Text, Image) and "Response" types and then have a union of the two 

but if there's no use for singling out the prompt types then we can keep as is

frontend/src/components/question-creation/question-elements/SaveQuestionEditorButton.tsx
joyce-shi: it has to be a question specifically! like they can add text / image wherever they want (i.e at the end)

frontend/src/components/assessments/assessment-creation/CreateAssessementButton.tsx
joyce-shi: you know what i think im dumb - i don't think we need the date in params at all, i think we can just call it inside the `CreateAssessmentHeader`

frontend/src/components/assessments/assessment-creation/CreateAssessmentHeader.tsx
```diff
@@ -4,31 +4,34 @@ import {
...
-  date,
   save,
 }: CreateAssessementHeaderProps): React.ReactElement => {
+  const date = new Date();
+  const options: Intl.DateTimeFormatOptions = {
+    month: "long",
+    day: "numeric",
+    year: "numeric",
+  };
+  const formattedDate = date.toLocaleDateString("en-US", options);
```
carissa-tang: can we move this date stuff to utils?

carissa-tang (commented):
just a side comment - i noticed the errors currently are attached to a given input. if that's intended behaviour, maybe it makes more sense to say smth like  "Please ensure this field is filled." or for error messages applicable in general maybe it should live at the top level
image

frontend/src/components/question-creation/question-elements/SaveQuestionEditorButton.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const validateNoExistingErrors = () => {
+    return questionElements.every(
+      (element) =>
+        !element.error ||
+        isQuestionPairError(element.error) ||
+        element.error === emptyElementError,
+    );
+  };
+
+  const validateNoEmptyEditorError = () => {
```
carissa-tang: i think this message is a little confusing bc its expecting specifically a question. maybe change the error message to "Please add at least one question element to the editor before saving" and rename `validateNoEmptyEditorError` -> `validateHasQuestionError` or smth

image

carissa-tang: also would we be able to remove the error message once a question is added (instead of only on `save`)

frontend/src/components/assessments/assessment-creation/CreateAssessmentHeader.tsx
joyce-shi: yah!

frontend/src/components/question-creation/question-elements/SaveQuestionEditorButton.tsx
joyce-shi: will do!

carissa-tang (approved):
LGTM!!

PR #182, opened 2023-03-30
feat: save assessment questions

joyce-shi:
## Notion ticket link

[Save Assessment Questions section on Create Assessment page](https://www.notion.so/uwblueprintexecs/Save-Assessment-Questions-section-on-Create-Assessment-page-a7843e084ec746ed9cfc2407b05f0e67?pvs=4)

## Implementation description
* Functionality for saving assessment questions

Note: I bumped the Typescript version to resolve an issue I was having defining optional array fields in interfaces, similar to the one described [here](https://github.com/microsoft/TypeScript/issues/41651).

## Steps to test
1. no breaking changes
2. elements of the question editor save properly


frontend/src/APIClients/mutations/TestMutations.ts
```diff
@@ -30,7 +30,7 @@ export const UNARCHIVE_TEST = gql`
   }
 `;
 
-export const SAVE_ASSESSMENT = gql`
+export const CREATE_NEW_ASSESSMENT = gql`
   mutation Create_Test($test: TestRequestDTO!) {
     createTest(test: $test) {
       id
```
joyce-shi, line 44: updated the naming, since this is the same mutation we should use when pressing the **Publish** button on the Create Assessment page

frontend/src/components/assessments/assessment-creation/BasicInformation.tsx
```diff
@@ -17,70 +19,90 @@ import {
...
-                  placeholder=""
+                  placeholder="Select a grade"
+                  selectedOptionStyle="check"
                   useBasicStyles
-                  value={value}
+                  value={
+                    gradeOptions.find(
+                      (option) => option.value === watch("grade"),
+                    ) || undefined
+                  }
```
joyce-shi, line 110: I think we should probably split some of these `Select` components into separate components. I think there's a couple bugs related to dropdowns and also a ticket for refactoring some stuff into common components, so I think I'll work on all of these things together in a subsequent PR! So leaving in this component for now

frontend/src/types/QuestionTypes.ts
```diff
@@ -1,10 +1,41 @@
...
-  MULTIPLE_CHOICE = "Multiple Choice",
-  SHORT_ANSWER = "Short Answer",
-  MULTI_SELECT = "Multi-select",
+  QUESTION_TEXT = "QUESTION_TEXT",
+  TEXT = "TEXT",
+  IMAGE = "IMAGE",
+  MULTIPLE_CHOICE = "MULTIPLE_CHOICE",
+  SHORT_ANSWER = "SHORT_ANSWER",
+  MULTI_SELECT = "MULTI_SELECT",
+}
```
joyce-shi, line 21: Updated the naming of this values of this enum to match the corresponding backend enum. Then, in all the places where the value was used (i.e as a text label), used our existing formatting utils.

frontend/src/types/QuestionTypes.ts
```diff
@@ -1,10 +1,41 @@
...
+
+export type QuestionElementDataType =
+  // remove string when all the elements have been implemented
```
joyce-shi, line 24: There's some placeholder values for other elements right now, so leaving this in. Later, when we have all the elements implemented, can remove this line!

frontend/src/utils/QuestionUtils.ts
```diff
@@ -98,7 +106,52 @@ export const getQuestionTexts = (question: QuestionElement[]): string[] => {
...
+};
+
+const formatMultipleChoiceRequest = (
+  data: MultipleChoiceData,
+): MultipleChoiceMetadata => {
+  return {
+    options: data.options.map((option) => option.value),
+    answerIndex: data.options.findIndex((option) => option.isCorrect),
+  };
+};
```
joyce-shi, line 125: for other question components, i just stored their `data` in their metadata types. but for multiple choice, since I needed an id field in its implementation, i just defined this extra helper function to convert it to its metadata type!

joyce-shi:
cc @cindy-531

package-lock.json
```diff
@@ -0,0 +1,15 @@
+{
```
carissa-tang: is this change needed ? since we are updating frontend/yarn.lock already

frontend/src/components/question-creation/question-elements/modals/multiple-choice/MultipleChoiceModal.tsx
```diff
@@ -33,7 +33,7 @@ const MultipleChoiceModal = ({
   const [emptyOptionError, setEmptyOptionError] = useState(false);
 
   useEffect(() => {
-    setOptionCount(data ? data.optionCount : 0);
+    setOptionCount(data ? data.options.length : 0);
```
carissa-tang: wait do we still need `optionCount` or can we just use `options.length`

carissa-tang (commented):
1. should we add a placeholder here
image
2. can we clear these errors on change
image

package-lock.json
joyce-shi: removed!

frontend/src/components/question-creation/question-elements/modals/multiple-choice/MultipleChoiceModal.tsx
joyce-shi: i think it still makes sense to keep `optionCount`? bc we use it as a local variable to add elements to `options` (so if we just used `options.length` it would always be 0)

joyce-shi:
addressed pr comments & added handling for multi-select (since jenna was able to get that merged yesterday!!) also removed `package-lock.json` since i think that was accidentally added

frontend/src/components/question-creation/question-elements/modals/multiple-choice/MultipleChoiceModal.tsx
carissa-tang: ah kk

carissa-tang (approved):
LGTM!!

PR #183, opened 2023-03-31
chore: add folder & renaming routes

joyce-shi:
## Implementation description
* Added `_PAGE` to the end of routes, to distinguish them from Queries and Mutations (since we use the same capitalization)
* Added in an `assessment-status` folder that I somehow missed in my last PR


carissa-tang (approved):
yay!

PR #184, opened 2023-04-01
backend: Remove Deleted Tests from Tests Query

wais-hundekar:
## Notion ticket link

[Remove Deleted Tests from Tests Query](https://www.notion.so/uwblueprintexecs/47bf4727a2444bccad4a91952898b8f3?v=9658436bf9be409e90581631a7155f16&p=2a46e3665b2d483aa781b640c2a5732a&pm=s)

## Implementation description
* Updated GetAllTests Query and anything else needed so that we can now query based on status
* Adding the status parameter is optional, so original functionality is maintained
* Enables future use cases where we only want to query tests of a certain status

## Steps to test
1. Visit Display Assessments Page and load tests
2. Change variables passed in to ensure functionality runs as intended, including passing in no variables at all

* All functionality runs as intended

## Checklist

### DIFF
--- a/backend/graphql/resolvers/testResolvers.ts
+++ b/backend/graphql/resolvers/testResolvers.ts
@@ -96,8 +96,11 @@ const testResolvers = {
     },
   },
   Query: {
-    tests: async (): Promise<TestResponseDTO[]> => {
-      return testService.getAllTests();
+    tests: async (
+      _: unknown,
+      { status }: { status?: AssessmentStatus[] },
+    ): Promise<TestResponseDTO[]> => {
+      return testService.getAllTests(status);
     },
   },
   Mutation: {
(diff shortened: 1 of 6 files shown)


joyce-shi:
I think because we're looking to update the function `getAllTests`, we don't want to have a status parameter (since then you wouldn't be getting **all** tests). If there is a use case for fetching by status, let me know and we can create a ticket for a separate function `getTestByStatus`

For this ticket, you should be able to just update `getAllTests` in `testService.ts` to filter out deleted tests.

joyce-shi (changes requested):
I think because we're looking to update the function getAllTests, we don't want to have a status parameter (since then you wouldn't be getting all tests). If there is a use case for fetching by status, let me know and we can create a ticket for a separate function getTestByStatus

For this ticket, you should be able to just update getAllTests in testService.ts to filter out deleted tests!

PR #189, opened 2023-04-03
fix assessments table sort

Nimasheth:
## Notion ticket link

[fix assessments table sort](https://www.notion.so/uwblueprintexecs/Fix-assessments-table-sort-d1773be45615493cb487c04716ad8631)

Default sort for Assessments Table should be set to Name, not Status

## Checklist

### DIFF
--- a/frontend/src/components/auth/Login.tsx
+++ b/frontend/src/components/auth/Login.tsx
@@ -32,6 +32,7 @@ const Login = (): React.ReactElement => {
 
   const onLogInClick = async () => {
     if (!(email && password)) {
+      setLoginError(true);
       return;
     }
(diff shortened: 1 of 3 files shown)


joyce-shi (changes requested):
hmm.. i think for the sort menu, the default sort should still be the first element, not the second. instead, would you be able to just switch the order of items in the assessment table sort? i.e name should come first, status should come second (and then default sort can stay as is)

Nimasheth:
> hmm.. i think for the sort menu, the default sort should still be the first element, not the second. instead, would you be able to just switch the order of items in the assessment table sort? i.e name should come first, status should come second (and then default sort can stay as is)

yep!!

joyce-shi (approved):
LGTM!!!

PR #195, opened 2023-04-08
feat: Image Upload Backend

carissa-tang:
## Notion ticket link

[Update backend to support Image Upload in Question Creation](https://www.notion.so/uwblueprintexecs/Update-backend-to-support-Image-Upload-in-Question-Creation-ac656aa3ff4c4409b897a841a51691c2?pvs=4)

## Implementation description
* Created `ImageStorageService` interface and implementation, and wrote unit tests
   * Most code pulled from `entityResolvers.ts` and `entityService.ts`
* Updated `TestService` to hydrate images in every API call
   * URL stored in `ImageMetadata` has an expiration time, so we need to make get request to get a new link that is guaranteed to be valid
      * Not sure if there's a more efficient way (e.g. only make get request if actually invalid) but I referenced Rowan-House's implementation for now
* Updated validators to support image validation

## Steps to test
1. yarn test

Some thoughts abt image expiration - rn by default it lives for 1hr. But we can customize it so that
   * URLs generated within question editor can live for 1hr (or less)
   * URLS generated on `createTest` / `updateTest` can live longer
      * I think max is 7 days, but I can see if there's another way to generate long-lived URLS

## Checklist


backend/services/implementations/testService.ts
```diff
@@ -1,22 +1,47 @@
...
+import IImageStorageService from "../interfaces/imageStorageService";
+import FileStorageService from "./fileStorageService";
+import ImageStorageService from "./imageStorageService";
 
 const Logger = logger(__filename);
 
 class TestService implements ITestService {
+  imageStorageService: IImageStorageService;
+
+  constructor() {
```
carissa-tang, line 24: not sure what is better practice - to initialize `fileStorageService` inside or pass as independent object and if it makes sense for test object to "own" images (which i think so?). but main reason was to reduce duplicated code

backend/services/interfaces/testService.ts
```diff
@@ -1,13 +1,13 @@
 import {
```
carissa-tang, line 1: renamed types for clarity

backend/middlewares/validators/util.ts
```diff
@@ -46,6 +48,10 @@ export const validateFileType = (mimetype: string): boolean => {
   return allowableContentTypes.has(mimetype);
 };
 
+export const validateImageType = (mimetype: string): boolean => {
```
joyce-shi, line 51: nit: mimetype to media type? https://en.wikipedia.org/wiki/Media_type

backend/services/implementations/imageStorageService.ts
```diff
@@ -0,0 +1,83 @@
...
+    } catch (error: unknown) {
+      Logger.error(
+        `Failed to create image. Reason = ${getErrorMessage(error)}`,
+      );
+      throw error;
+    }
+  }
+}
+
+export default ImageStorageService;
```
joyce-shi: nit: but should this be called `imageUploadService` instead? if i'm understanding correctly, it seems like this service handles the uploading logic and wraps `fileStorage` handle the actual storage

backend/services/implementations/testService.ts
joyce-shi: i think it makes sense for each test object to own its own images, but i think if you're not using `fileStorageService` inside `testService`, it probably makes sense to initialize it within `imageStorageService`, unless there's something i'm missing!

backend/models/test.model.ts
```diff
@@ -51,7 +51,8 @@ export interface TextMetadata {
  * This interface contains additional information about an image component
  */
 export interface ImageMetadata {
-  src: string;
+  url: string;
+  filePath: string;
```
joyce-shi, line 55: why do we need both `url` and `filepath`? sure you have your reasons, but curious!

joyce-shi (approved):
LGTM! left some minor comments

backend/middlewares/validators/util.ts
carissa-tang: does it make sense to keep the term `mimetype` since graphql's `FileUpload` uses that naming?

backend/models/test.model.ts
carissa-tang: `url` is the "preview link" we can use to display the photo but has expiration time
`filePath` persists (has no expiration) and is used to generate new url for the photo

carissa-tang: actually with the way it currently works, we don't really need to store the url cause we regenerate it every time

carissa-tang: do you think makes more sense to just store the `filePath`? and then return only the `url`

backend/middlewares/validators/util.ts
joyce-shi: oh probably yes then LOL nvmd

backend/models/test.model.ts
joyce-shi: i think that makes more sense to me! but if it's a lot of extra effort it's probably fine as is

carissa-tang: i think i'm gonna leave as is actually bc it gets a bit ugly converting between the two types - we can refactor later if needed

PR #196, opened 2023-04-09
feat: create classroom using mutation from create classroom modal

d282liu:
## Notion ticket link
[Create mutation to Create Classroom](https://www.notion.so/uwblueprintexecs/Create-mutation-to-Create-Classroom-16dd181c8ab8492c9ec3ac8f26f2141c?pvs=4)

## Implementation description
* Created `frontend/src/APIClients/mutations/ClassMutations.ts` to call `createClass` mutation from frontend
* Made changes to classroom modal and related front end components to call `createClass` mutation

## Steps to test
1. Test front end 
2. Check MongoDB atlas when saving classes to make sure all the fields are filled in properly

## Checklist


frontend/src/components/user-management/student/AddClassroomConfirmationMessage.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
```
carissa-tang: i believe there should be a toast and not this confirmation message

image

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -34,13 +41,22 @@ const AddClassroomModal = (): React.ReactElement => {
     formState: { errors },
   } = useFormContext<ClassroomForm>();
   const { onOpen, onClose, isOpen } = useDisclosure();
+  const [className, setClassName] = useState("");
+  const [schoolYear, setSchoolYear] = useState(0);
+  const [gradeLevel, setGradeLevel] = useState(Grade.K);
+  const [teacher, setTeacher] = useState("6347056d47cd96025c18e639");
```
carissa-tang:
```suggestion
```
we can directly pass the id from AuthContext (and so we don't need useState)

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -34,13 +41,22 @@ const AddClassroomModal = (): React.ReactElement => {
...
+  const [testSessions, setTestSessions] = useState([]);
```
carissa-tang:
```suggestion
```
 i missed this as well from last PR, but can we remove testSessions from `ClassRequestDTO`. it is always empty during creation and we can't update it in the edit classroom modal. You will need to update BE and initialize field to [] in `createClass`

we will handle the logic for test sessions later on

carissa-tang: @joyce-shi pls correct me if i'm missing smth

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -102,7 +123,27 @@ const AddClassroomModal = (): React.ReactElement => {
       setRequestErrorMessage(
         "Please ensure all required components are filled out before submitting your application.",
       );
+      return;
     }
+    const classObj: ClassRequest = {
```
carissa-tang: can we use `handleSubmit` here? then we can use the `data` object instead of having all the useState fields (example code [here](https://github.com/uwblueprint/jump-math/blob/ca65dc5688234aa8e037f89b29f149db32c834f2/frontend/src/components/pages/admin/CreateAssessmentPage.tsx#L52))

carissa-tang (commented):
sorry i missed this earlier - the `gradeLevel` [field](https://github.com/uwblueprint/jump-math/blob/staging/backend/models/class.model.ts#L55) should be just a string (not an array of strings). can you update that in the backend everwhere?

d282liu:
Hello, I made the changes! One thing - I noticed in the design that after the initial save, the modal changes the buttons from 'discard' and 'save' to say 'back' and 'confirm' respectively, and when hitting back, it returns you to the previous screen. I feel this is a bit janky, since the modal doesn't change except for the buttons - it would make more sense if there was an actual confirmation page listing out the fields or something like that. For now, I made the change to show the warning upon the first save click, but the buttons do not change text because I have not figured out how to do that just yet LOL and hitting discard just closes the modal rather than going back. Let me know what you think and I will make the required changes @carissa-tang

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -88,23 +100,61 @@ const AddClassroomModal = (): React.ReactElement => {
...
+      setShowWarning(false);
       setRequestErrorMessage(
-        "Please ensure all required components are filled out before submitting your application.",
+        "Please ensure all required components are filled out before saving changes",
       );
+    } else if (!showWarning) {
+      setShowWarning(true);
+      setShowRequestError(false);
+    } else {
+      console.log(`Classname: ${data.className}`);
```
carissa-tang: can we remove all the console logs!

frontend/src/APIClients/mutations/ClassMutations.ts
```diff
@@ -0,0 +1,19 @@
...
+    createClass(classObj: $classObj) {
+      id
+      className
+      schoolYear
+      gradeLevel
+      teacher {
+        id
+        firstName
+        lastName
+      }
```
carissa-tang:
```suggestion
```
i don't think we'll be needing this

frontend/src/components/common/WarningToast.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Alert, AlertIcon } from "@chakra-ui/react";
+
+interface WarningToastProps {
+  warningMessage: string;
+}
+
+const ErrorToast = ({
```
carissa-tang:
```suggestion
const WarningToast = ({
```

frontend/src/components/user-management/student/SelectFormInputClassroom.tsx
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import { UseFormSetValue, UseFormWatch } from "react-hook-form";
 import { GroupBase, Select, SingleValue } from "chakra-react-select";
 
+import { Grade } from "../../../APIClients/types/UserClientTypes";
```
carissa-tang: is this necessary?

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -18,29 +19,41 @@ import {
...
     setValue,
     formState: { errors },
   } = useFormContext<ClassroomForm>();
+  const { authenticatedUser } = useContext(AuthContext);
   const { onOpen, onClose, isOpen } = useDisclosure();
   const [classNameError, setClassNameError] = React.useState(false);
   const [schoolYearError, setSchoolYearError] = React.useState(false);
   const [gradeLevelError, setGradeLevelError] = React.useState(false);
   const [showRequestError, setShowRequestError] = useState(false);
+  const [showWarning, setShowWarning] = useState(false);
```
carissa-tang: as joyce mentioned we don't need the warning message! sorry :/

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -18,21 +19,27 @@ import {
...
 const AddClassroomModal = (): React.ReactElement => {
   const {
+    handleSubmit,
     watch,
     setValue,
     formState: { errors },
   } = useFormContext<ClassroomForm>();
+  const { authenticatedUser } = useContext(AuthContext);
   const { onOpen, onClose, isOpen } = useDisclosure();
   const [classNameError, setClassNameError] = React.useState(false);
```
carissa-tang: can we remove these 3 errors and the associated lines (61-74, 79, 84, 89)? i don't think we actually need them

backend/testUtils/class.ts
```diff
@@ -110,7 +105,6 @@ export const assertResponseMatchesExpected = (
   expect(result.schoolYear).toEqual(expected.schoolYear);
   expect(result.gradeLevel.toString).toEqual(expected.gradeLevel.toString);
   expect(result.teacher).toEqual(mockTeacher);
-  expect(result.testSessions).toEqual([mockTestSessionWithId]);
```
carissa-tang, line 113: i think we still need this check. i would suggest you create new constants with the expected values for each call

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -92,19 +101,41 @@ const AddClassroomModal = (): React.ReactElement => {
...
+        variables: {
+          classObj: {
+            ...data,
+            schoolYear: parseInt(data.schoolYear, 10),
+            teacher: authenticatedUser?.id,
+          },
+        },
+      })
+        .then(() => {
+          if (showRequestError) setShowRequestError(false);
```
carissa-tang:
```suggestion
```

frontend/src/components/user-management/student/AddClassroomModal.tsx
```diff
@@ -92,19 +101,41 @@ const AddClassroomModal = (): React.ReactElement => {
...
-      console.log(`Grade Level: ${watch("gradeLevel")}`);
-    } else {
+  const onSave: SubmitHandler<ClassroomForm> = async (data) => {
+    if (!validateFields()) {
       setShowRequestError(true);
       setRequestErrorMessage(
-        "Please ensure all required components are filled out before submitting your application.",
+        "Please ensure all required components are filled out before saving changes",
       );
+    } else {
```
carissa-tang, line 90:
```suggestion
    } else {
    if (showRequestError) setShowRequestError(false);
```

carissa-tang (commented):
line 178 should still be `type="number"` since we don't want to allow alpha characters, and you can set default value to 2023.

backend/testUtils/class.ts
d282liu: I just added the assertion to assert that `result.testSession` is equal to `[]`, is this okay? None of the tests acutally update the class to hold test sessions, since `createClass` and `updateClass` take in ClassRequestDTOs which no longer carry the `testSessions` field, I am assuming in the future there will be additional functions to add testSessions which will then be more properly tested, like `createStudent`, `update_student` and such. Let me know if this assertion should be changed

carissa-tang: wait but `getClassById` for example will still return test sessions, so that unit test should use a mockClass that has test sessions. whereas for create and edit, the mock value used will not have that field

d282liu: Ah yes this is true - I made a change to assert that specifically in getClassById test, let me know if that works

carissa-tang: hmm, sorry for being picky here.. but just to avoid more refactoring in the future since we'll be adding new functions that do have test sessions i think it might be better to directly use the `assertResponseMatchesExpected` function. i believe you can still do comparison if the value is undefined - if not, you can add a check if expected.testSessions is defined

d282liu: Oh whoops I am just seeing this now - I made the change, let me know if this works

carissa-tang (approved):
lgtm!!

PR #198, opened 2023-04-10
fix: teacher signup back button

Nimasheth:
## Notion ticket link

[Fix teacher signup back button](https://www.notion.so/uwblueprintexecs/Fix-teacher-signup-back-button-eb564f02bc854ce0b2d9054250e16601)

Teacher signup page 4 back button goes to page 3 instead of page 2

It should go back to page 2

## Checklist


carissa-tang (commented):
actually we do want to go to page 3 if they came from that page
- a hacky solution you can use is just to check if `school.name` field is not empty (that means they came from page 3) and return to page 3. otherwise, return to page 2

carissa-tang (approved):
lgtm!!

PR #203, opened 2023-04-11
fix: disappearing confirmation modal

joyce-shi:
## Notion ticket link

[Fix disappearing confirmation modal](https://www.notion.so/uwblueprintexecs/Fix-disappearing-confirmation-modal-a115b45dfc614591bcf0384c41b91f5b?pvs=4)

## Implementation description
* Instead of showing a successfully removed message, showing a Toast message to match the implementation for assessment table updates

* Works as described

### DIFF
--- a/frontend/src/components/user-management/RemoveUserModal.tsx
+++ b/frontend/src/components/user-management/RemoveUserModal.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { useHistory } from "react-router-dom";
 import { useMutation } from "@apollo/client";
 import { Button, useDisclosure } from "@chakra-ui/react";
 
@@ -9,8 +8,8 @@ import {
   GET_USERS_BY_ROLE,
 } from "../../APIClients/queries/UserQueries";
 import { CloseOutlineIcon } from "../../assets/icons";
-import { USER_DATABASE_PAGE } from "../../constants/Routes";
 import Modal from "../common/Modal";
+import Toast from "../common/Toast";
 
 interface RemoveUserModalProps {
   name: string;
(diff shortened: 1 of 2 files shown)


frontend/src/components/user-management/RemoveUserModal.tsx
```diff
@@ -39,41 +33,34 @@ const RemoveUserModal = ({
     },
   );
 
+  const { showToast } = Toast();
```
carissa-tang, line 36: can we add a bit extra spacing between the top of the screen and the toast?

joyce-shi: the toast component from chakra has preset position [options](https://chakra-ui.com/docs/components/toast/usage#changing-the-toast-position) and i don't think you can customize those. if you want to increase spacing, i think you have to build a custom toast component (i.e from `<Box>`). do you want me to do that?

carissa-tang: does it work if you update the custom theme? smth like
```
const Alert = {
  baseStyle: {
    container: {
      ...style...
    }
  }
}
```

joyce-shi: yes. LOL

frontend/src/themes/components/alert.ts
```diff
@@ -40,13 +40,15 @@ const Alert = {
     },
     successToast: {
       container: {
+        margin: "4",
```
carissa-tang, line 43: do we want this margin thing in `baseStyle` so it applies to all variants? not sure what the other two are tho / where it's used

joyce-shi: toast uses alert under the hood, so i don't think so? bc we don't want to apply the margin to other non-toast alerts (i.e form error messages). let me know if i'm missing something b/c i am not on it today

carissa-tang: ok ya that makes sense

carissa-tang (approved):

PR #205, opened 2023-04-12
fix: Prevent user self-deletion

JennaXiao3:
## Notion ticket link

[Prevent user self-deletion](https://www.notion.so/uwblueprintexecs/Prevent-user-self-deletion-3dd48272e74a4606aea7bc2239bb5640?pvs=4)

## Implementation description
* checked current user to determine if the three dots beside the admin in admin user table will display

## Steps to test
1. [localhost](http://localhost:3000/admin/user-database) 
2. find your email login
3. check if the three dots are at the right

* 

notes: the box with height 32px is to replace the space that would've been taken by the 3 dots -> or else it will be shorter in height than the other rows

## Checklist


carissa-tang (approved):
lgtm!!

PR #211, opened 2023-04-16
backend: Add Fraction Component 

cindy-531:
## Notion ticket link

[Update backend to Support Fraction Input](https://www.notion.so/uwblueprintexecs/Update-Backend-to-Support-Fraction-Input-eafe193c4dea42e39a3d4b8afa48a6c2?pvs=4)

## Implementation description
* Changed backend files to support the fraction question component type

## Steps to test
1. Run yarn test in /backend

* 

## Checklist

### DIFF
--- a/backend/graphql/types/testType.ts
+++ b/backend/graphql/types/testType.ts
@@ -63,22 +63,22 @@ const testType = gql`
 
   input MultipleChoiceMetadataInput {
     options: [String!]!
-    answerIndex: Float!
+    answerIndex: Int!
   }
 
   type MultipleChoiceMetadata {
     options: [String!]!
-    answerIndex: Float!
+    answerIndex: Int!
   }
 
   input MultiSelectMetadataInput {
     options: [String!]!
-    answerIndices: [Float!]!
+    answerIndices: [Int!]!
   }
 
   type MultiSelectMetadata {
     options: [String!]!
-    answerIndices: [Float!]!
+    answerIndices: [Int!]!
   }
 
   input ShortAnswerMetadataInput {
--- a/backend/graphql/types/testType.ts
+++ b/backend/graphql/types/testType.ts
@@ -8,6 +8,7 @@ const testType = gql`
     MULTIPLE_CHOICE
     MULTI_SELECT
     SHORT_ANSWER
+    FRACTION
   }
 
   enum AssessmentTypeEnum {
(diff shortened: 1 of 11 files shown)


backend/services/implementations/testSessionService.ts
```diff
@@ -352,7 +353,12 @@ class TestSessionService implements ITestSessionService {
     // - index (for multiple choice)
     // - list of indices (for multiple select)
     // - null (for no answer)
-    const studentTestAnswers: (number[] | number | null)[][] = result.answers;
+    const studentTestAnswers: (
+      | number[]
+      | number
+      | null
+      | FractionMetadata
```
joyce-shi: i'm not sure it makes sense here to just tack on another type. instead, it might make more sense to convert the `answers` field type in `ResultRequestDTO` from `(number[] | number | null)[][]` to `(number[] | null)` and for short answer / multiple choice responses, just store as an array of size 1. i think then we can combine the `getCorrectAnswer` and `getCorrectAnswers` functions  here for a cleaner implementation (and also keep the typing a bit cleaner)

cc @carissa-tang since you have a bit more context, is there any reason why this wouldn't work? i also think should be a minor change, but lmk if i'm wrong about that and if we should push that work to another PR.
ik you also have a ticket coming up to refactor the test session grading, so lmk if it makes more sense to do it there.

joyce-shi (commented):
great work! left a minor comment (we can wait on carissa's input to see if the change is necessary)

one thing - i noticed in testing section of your PR, you mentioned running `yarn test`, but were you also able to run create / update / get functions in the graphQL playground for the fraction input to verify that they work as expected?

backend/middlewares/validators/testValidators.ts
```diff
@@ -113,6 +114,23 @@ export const questionsValidator = (
         if (!validatePrimitive(questionComponent.metadata.answer, "number")) {
           throw new Error("The answer field is not of type number");
         }
+      } else if (questionComponent.type === QuestionComponentType.FRACTION) {
+        if (!("numerator" in questionComponent.metadata)) {
+          throw new Error("Fraction component is missing a numerator field");
+        }
+        if (
+          !validatePrimitive(questionComponent.metadata.numerator, "number")
```
carissa-tang:
```suggestion
          !validatePrimitive(questionComponent.metadata.numerator, "integer")
```
can you make this change for both fields? also for line 82 and 104

backend/graphql/types/testType.ts
```diff
@@ -88,13 +89,24 @@ const testType = gql`
     answer: Float!
   }
 
+  input FractionMetadataInput {
```
carissa-tang, line 92: could you also fix the multiple choice and multi select to use Int for the indices 🙏 TYSM

backend/services/implementations/testSessionService.ts
carissa-tang: we can make the refactor in a separate PR. for here, we can store the fraction answer as [numerator, denominator] ? (and keep the fractionMetadata object)

carissa-tang (approved):
LGTM!!

PR #213, opened 2023-04-17
feat: student dashboard routing

joyce-shi:
## Implementation description
* linked together student dashboard pre-assessment pages
* set up routing for student dashboard
* set up context for student authentication (i'll populate this in a subsequent PR after the necessary backend work is completed)
Note: I haven't created the context that will wrap the `StudentDashboard` in this PR (i.e with the test info), since I think it makes more sense to do that when I actually query for test data. i'll do that in a follow-up PR! in this PR, mainly just putting together some of the completed frontend work for the student dashboard 

* no breaking changes (in particular due to the `AuthTypes` changes)
* can only navigate to `StudentDashboard` pages when "authenticated" (i.e put in the passcode & selected a student)

### DIFF
--- a/frontend/src/components/auth/student-login/NameSelection.tsx
+++ b/frontend/src/components/auth/student-login/NameSelection.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useContext } from "react";
 import { Controller, useForm } from "react-hook-form";
 import { useHistory } from "react-router-dom";
 import {
@@ -10,9 +10,10 @@ import {
 import { Select } from "chakra-react-select";
 
 import { STUDENT_SIGNUP_IMAGE } from "../../../assets/images";
-import { HOME_PAGE } from "../../../constants/Routes";
-import AuthWrapper from "../../auth/AuthWrapper";
-import NavigationButtons from "../../auth/teacher-signup/NavigationButtons";
+import { ASSESSMENT_SUMMARY_PAGE, HOME_PAGE } from "../../../constants/Routes";
+import AuthContext from "../../../contexts/AuthContext";
+import AuthWrapper from "../AuthWrapper";
+import NavigationButtons from "../teacher-signup/NavigationButtons";
 
 const students = [
   {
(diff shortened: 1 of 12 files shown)


frontend/src/components/auth/student-login/NameSelection.tsx
```diff
@@ -89,7 +91,16 @@ const StudentNameSelection = (): React.ReactElement => {
...
-        onContinueClick={() => undefined}
+        onContinueClick={() => {
+          setAuthenticatedUser({
+            id: "temporary",
+            firstName: "temporary",
+            lastName: "temporary",
+            role: "Student",
+            testId: "temporary",
+          });
+          history.push(ASSESSMENT_SUMMARY_PAGE);
```
joyce-shi, line 102: I'll cleanup this logic once i update the backend logic and fetch the data!

frontend/src/components/pages/student/StudentDashboard.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
carissa-tang, line 1: does this mean once they're logged in they'll have access to the test forever? since we don't check validity of test session beyond the access code part

carissa-tang (commented):
just one question!

frontend/src/components/pages/student/StudentDashboard.tsx
joyce-shi: a couple things here:
- whenever the browser is refreshed, the context should automatically clear, since we don't persist the student auth values to local storage). so theoretically, the user won't be able to access the assessment anymore (let me know if that seems to not be the case)
- there's a ticket in the current sprint to update the access code values so that after the time has expired, student users also shouldn't be able to access test sessions
let me know if that answers your question!

carissa-tang: is that validation happening in `AssessmentSummaryPage`? i.e. if they stay on that page beyond the expiration time , will they be kicked out

carissa-tang (approved):

frontend/src/components/pages/student/StudentDashboard.tsx
joyce-shi: that's a good point. i think theoretically if they never closed their browser they could see the test forever? but yeah, i can work on adding some sort of session expiry in a future ticket

joyce-shi: https://www.notion.so/uwblueprintexecs/Student-Dashboard-session-expiration-000487c0e65d4210964813d5af68555c?pvs=4

PR #214, opened 2023-04-20
feat: populate name selection page

joyce-shi:
## Notion ticket link

[Fetch data for Name Selection page](https://www.notion.so/uwblueprintexecs/Fetch-data-for-Name-Selection-page-833508229cf3431d8f7aa20d1523518c?pvs=4)
[Update backend to fetch data for Name Selection page](https://www.notion.so/uwblueprintexecs/Update-backend-to-fetch-data-for-Name-Selection-page-3c1ac2535df44a45aeb1c9664d63a5f6?pvs=4)

## Implementation description
* Implemented `getClassByTestSessionId` logic
* Populated Name Selection page using `getClassByTestSessionId`
* Stored results in AuthContext for future use

* Data is populated correctly
* No breaking changes


backend/services/implementations/testService.ts
```diff
@@ -166,7 +166,7 @@ class TestService implements ITestService {
         throw new Error(`Test ID ${id} not found`);
       }
       // eslint-disable-next-line no-underscore-dangle
-      test._id = mongoose.Types.ObjectId();
+      test._id = new mongoose.Types.ObjectId();
```
joyce-shi, line 169: compiler was throwing an error

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -106,6 +106,35 @@ describe("mongo classService", (): void => {
...
+    await ClassModel.create(testClass[0]);
+    await ClassModel.create(testClass[0]);
+    const testSessionId = testClass[0].testSessions[0];
+    await expect(async () => {
+      await classService.getClassByTestSessionId(testSessionId);
+    }).rejects.toThrowError(
+      `More than one class has the same Test Session of id ${testSessionId}`,
+    );
+  }); */
+
```
joyce-shi, line 136: originally, i had created these tests before david removed adding test sessions from the `createClass` logic. but now that the logic for adding test sessions to classes doesn't exist yet, i don't think there's a way to write unit tests for the `getClassByTestSessionId` function and i think adding that logic is beyond the scope of this PR!

so leaving tests off of this pr and  created a separate [ticket](https://www.notion.so/uwblueprintexecs/Update-backend-to-support-distributing-assessment-7efdcc435f2c46a097cb678839139a07?pvs=4) to track that work. I also left a note to uncomment these tests (and get them working) after this functionality is added!

backend/services/implementations/classService.ts
```diff
@@ -72,6 +72,28 @@ class ClassService implements IClassService {
...
+          `More than one class has the same Test Session of id ${testSessionId}`,
+        );
+      }
+    } catch (error: unknown) {
+      Logger.error(`Failed to get Class. Reason = ${getErrorMessage(error)}`);
+      throw error;
+    }
+    return (await this.mapClassToClassDTOs(classes))[0];
+  }
+
```
joyce-shi, line 96: created a [ticket](https://www.notion.so/uwblueprintexecs/Update-getClassByTestSession-to-support-completed-tests-852efb2999cb4033884362b9993605e4?pvs=4) on the board to update this function with handling to filter out students who have already taken the test. not adding this logic now, since i think it makes sense to do that after we implement test submission logic! let me know if you think otherwise and i can add that

frontend/src/APIClients/queries/TestSessionQueries.ts
```diff
@@ -4,6 +4,9 @@ const GET_TEST_SESSION_BY_ACCESS_CODE = gql`
   query TestSessionByAccessCode($accessCode: String!) {
     testSessionByAccessCode(accessCode: $accessCode) {
       id
+      test {
+        id
+      }
```
joyce-shi, line 9: so i realized we actually store the entire test object inside a test session. i don't think we need to do that since we should probably need to make a separate query for the test data in the `AssessmentSummary` page (i.e after student authentication is complete). given this, should i make a PR to only store the `test_id` in a `test_session` object? or is there another purpose for this

frontend/src/components/auth/student-login/NameSelection.tsx
```diff
@@ -1,105 +1,97 @@
...
+  testSessionId: string;
+}
+
+interface StudentOption extends OptionBase {
+  value: StudentResponse;
+  label: string;
+}
+
+const NameSelection = ({
+  testId,
```
joyce-shi, line 31: the `testId` value is currently unused, since we don't need it here. i was thinking that i would pass it via props to the `AssessmentSummaryPage` where it could query for test data and then store the results in a context that wraps the `StudentDashboard`? and then `AuthContext` only stores info related to the student specifically.

backend/models/class.model.ts
```diff
@@ -22,9 +22,13 @@ export interface Class extends Document {
 }
 
 export interface Student {
+  /** the unique identifier for the student */
   id: string;
+  /** the first name of the student */
   firstName: string;
+  /** the last name of the student */
   lastName: string;
+  /** an optional identifier provided by the teacher */
```
carissa-tang, line 31: is this unique as well?

backend/services/interfaces/classService.ts
```diff
@@ -74,7 +82,7 @@ export interface IClassService {
    */
   createStudent(
     student: StudentRequestDTO,
-    classId: string,
+    id: string,
```
carissa-tang: i think `id` might be confused with `studentId` instead of class? but if you think this is more clear then we can make the change!

backend/graphql/types/classType.ts
```diff
@@ -31,6 +31,10 @@ const classType = gql`
     students: [StudentResponseDTO]!
   }
 
+  extend type Query {
+    classByTestSession(testSessionId: String!): ClassResponseDTO!
```
carissa-tang:
```suggestion
    classByTestSession(testSessionId: ID!): ClassResponseDTO!
```

backend/services/implementations/__tests__/classService.test.ts
carissa-tang: wait i'm confused - can we use [testClassWithTestSessions](https://github.com/uwblueprint/jump-math/blob/staging/backend/testUtils/class.ts#L74) for these tests? or is there something else missing

frontend/src/APIClients/queries/TestSessionQueries.ts
carissa-tang: hmm yea, make sense to have a separate query (can't think of a use case off the top of my head). we should probably re-evaluate our other functions as well (i.e. if we need the ResponseDTOs or just the id)

backend/services/implementations/__tests__/classService.test.ts
joyce-shi: wait i'm not sure if i'm misunderstanding, b/c i see one test case that uses `testClassWithTestSessions`. but didn't we remove the logic to include test sessions in `createClass`? so, how would do we store the `testSession` in the RequestDTO?

carissa-tang: the `createClass` we wrote doesn't support adding test sessions but in the unit test you can use Model.create() to still create an object with test sessions (to mimick having called createClass + addTestSessionToClass). not sure if that answers your question

carissa-tang: the test case that uses it is a getter, so we would expect some objects in our database to already have test sessions

joyce-shi: okay sounds good! thank you (:

backend/models/class.model.ts
joyce-shi: uhhh sort of? I guess it would probably be a unique input provided by the teacher (i.e no two students in the class would have the same identifier), but there is probably no guarantee of "uniqueness" across our system, since theoretically different teachers could have the same identifiers for different students (i.e if they number them #1, #2, etc.). so i hesitate to call it a `unique identifier`, but let me know if there is some way i can update the documentation to be more clear

frontend/src/APIClients/queries/TestSessionQueries.ts
joyce-shi: makes sense. added a [ticket](https://www.notion.so/uwblueprintexecs/Refactor-Test-Session-model-to-omit-Test-object-91e7fabcbc464fec8b6cbaefac910353?pvs=4) for storing only the id! i'll work on it after we finish the backend changes for the teacher and student dashboards, in case we do end up needing it.

for the other functions, agree. maybe we can create tickets as we complete the backend work coming up and then at the end, we can do a full cleanup?

carissa-tang: sg!

backend/models/class.model.ts
carissa-tang: no that makes sense that we can't guarantee uniqueness! thanks for clarifying

carissa-tang (approved):
lgtm! 

side note: i noticed that the teacher ids in Class and Test Session do not agree. prob just a result of manually adding ids? should be resolved when we clean up DB but ig just smth to keep in mind in case it causes weird behaviour.

PR #226, opened 2023-05-01
backend: testSessionByTeacherId Resolver and Query

wais-hundekar:
## Notion ticket link

Update Backend to Fetch Test Sessions by Teacher](https://www.notion.so/uwblueprintexecs/Update-backend-to-fetch-test-sessions-by-teacher-5e3c9d2e2ef649759390dd61b7fe715e)

## Implementation description
* Implemented Resolver and Query for testSessionByTeacherId
Screenshot 2023-05-01 at 1 14 02 AM
Screenshot 2023-05-01 at 7 54 00 PM

## Steps to test
1. Test query with command in screenshot

* Ensure all types are correct
* Ensure that query return is also correct

## Checklist

### DIFF
--- a/backend/graphql/resolvers/testSessionResolvers.ts
+++ b/backend/graphql/resolvers/testSessionResolvers.ts
@@ -59,6 +59,12 @@ const testSessionResolvers = {
     ): Promise<TestSessionResponseDTO> => {
       return testSessionService.getTestSessionByAccessCode(accessCode);
     },
+    testSessionByTeacherId: async (
+      _parent: undefined,
+      { teacherId }: { teacherId: string },
+    ): Promise<Array<TestSessionResponseDTO>> => {
+      return testSessionService.getTestSessionsByTeacherId(teacherId);
+    },
   },
   Mutation: {
     createTestSession: async (
--- a/backend/graphql/resolvers/testSessionResolvers.ts
+++ b/backend/graphql/resolvers/testSessionResolvers.ts
@@ -59,7 +59,7 @@ const testSessionResolvers = {
     ): Promise<TestSessionResponseDTO> => {
       return testSessionService.getTestSessionByAccessCode(accessCode);
     },
-    testSessionByTeacherId: async (
+    testSessionsByTeacherId: async (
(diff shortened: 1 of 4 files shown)


backend/graphql/resolvers/testSessionResolvers.ts
```diff
@@ -59,6 +59,12 @@ const testSessionResolvers = {
     ): Promise<TestSessionResponseDTO> => {
       return testSessionService.getTestSessionByAccessCode(accessCode);
     },
+    testSessionByTeacherId: async (
```
carissa-tang:
```suggestion
    testSessionsByTeacherId: async (
```

carissa-tang (commented):
can you test what happens if you call this query on a teacher with 0 test sessions? would be great if you can add a unit test for that case. you might also need to update the return type since might be []

wais-hundekar:
> can you test what happens if you call this query on a teacher with 0 test sessions? would be great if you can add a unit test for that case. you might also need to update the return type since might be []

@carissa-tang What would be the difference between this and the existing "getTestSessionsByTeacherId for invalid teacher id" test? If the function works by searching through all the test sessions to check if the teacherId matches for any of them, wouldn't the invalid id case also be the same result as the case where no matching teacher id exists (no test session for a teacher?)

frontend/src/APIClients/queries/TestSessionQueries.ts
```diff
@@ -11,4 +11,13 @@ const GET_TEST_SESSION_BY_ACCESS_CODE = gql`
   }
 `;
 
-export default GET_TEST_SESSION_BY_ACCESS_CODE;
+export const GET_TEST_SESSION_BY_TEACHER_ID = gql`
+  query TestSessionByTeacherId($teacherId: String!) {
+    testSessionByTeacherId(teacherId: $teacherId) {
```
carissa-tang:
```suggestion
    testSessionsByTeacherId(teacherId: $teacherId) {
```

carissa-tang (approved):

PR #228, opened 2023-05-01
feat: Update createTestSession to store reference in related class

carissa-tang:
## Notion ticket link

[Update createTestSession logic to store reference in related class](https://www.notion.so/uwblueprintexecs/Update-createTestSession-logic-to-store-reference-in-related-class-97096434af6b43179bc4310f9cb04bb3?pvs=4)

## Implementation description
* I added the function directly in TestSessionService, which is a bit not ideal since it directly uses MgClass, but the two services would otherwise have a circular dependency. And I don't think we need to store a class service within test session

## Steps to test
1. yarn test
2. GraphQL

image
(second test session)
image

* 

## Checklist

### DIFF
--- a/backend/services/interfaces/testSessionService.ts
+++ b/backend/services/interfaces/testSessionService.ts
@@ -110,11 +110,13 @@ export interface ResultResponseDTO {
 export interface ITestSessionService {
   /**
    * create a TestSession with the fields given in the DTO, return created TestSession
+   * @param classId id of the class to create the TestSession in
    * @param testSession new testSession
    * @returns the created TestSession
    * @throws Error if creation fails
    */
   createTestSession(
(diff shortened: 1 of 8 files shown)


backend/services/interfaces/testSessionService.ts
```diff
@@ -110,11 +110,13 @@ export interface ResultResponseDTO {
 export interface ITestSessionService {
   /**
    * create a TestSession with the fields given in the DTO, return created TestSession
+   * @param classId id of the class to create the TestSession in
```
joyce-shi:
```suggestion
   * @param id of the class taking the test session
```

joyce-shi (approved):
LGTM!

PR #232, opened 2023-05-01
feat: Student Dashboard Header

JennaXiao3:
## Notion ticket link

[Student Dashboard Header](https://www.notion.so/uwblueprintexecs/Student-Dashboard-Header-2921ab2ffb2540c7be24ad40bd40f9ea?pvs=4)

## Implementation description
* modified Navbar to incorporate the header (since there are some similarities such as the HomeButton and bottom border)
* added another type (StudentDashboardInfoTypes)

## Steps to test
1. go to [localhost](http://localhost:3000/component-library)

* should the estimated time be some type of time object or is string fine?
* a separate component was initially made for StudentDashboardHeader but ended up modifying Navbar instead since it shared similarities...is this okay?

## Checklist


joyce-shi:
i agree that there are similarities between the existing navbar and this new navbar for the student experience - good catch noticing this!

however, i think rather than having "2 versions" of the navbar in one component, a potential cleaner approach may be to create a wrapper component from the existing implementation and then having 2 child components (one for the existing one vs. one for the new ones), so that we can keep our components to [single purpose](https://blog.bitsrc.io/single-responsibility-principle-for-react-js-applications-5cde2fd49f11). let me know what you think of this!

JennaXiao3:
> single purpose

yeah this sounds like a much better approach i'll do that!

frontend/src/components/assessments/assessment-creation/StudentDashboardHeader.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Box, HStack, Text, VStack } from "@chakra-ui/react";
+
+import HeaderWrapper from "../../common/HeaderWrapper";
+
+interface StudentDashboardHeaderProps {
+  assessmentName: string;
+  classroomName: string;
+  estimatedTime: string;
```
joyce-shi: can you check w/ design if `estimatedTime` is still needed in the header? i think they might have made some changes since the ticket was created.

riishi is the designer to reach out to!

frontend/src/components/common/HeaderWrapper.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Box, Flex } from "@chakra-ui/react";
+
+import HomeButton from "./HomeButton";
+
+interface HeaderWrapperProps {
+  content: React.ReactElement;
+}
+
+const HeaderWrapper = ({ content }: HeaderWrapperProps): React.ReactElement => {
```
joyce-shi: instead of taking the `content` in as props, you can just use React `children`! there are existing examples in the codebase on how to implement this. you can checkout `QuestionSummary.tsx` and `AssessmentInfo.tsx` for one such example.

it's basically what you are implementing here, except it let's you wrap your the content components (it'll make more sense if you look at the codebase).

joyce-shi (commented):
looks great! left some minor comments (: also i think the build is failing rn, you might need to pull latest master to fix it!

frontend/src/components/common/HeaderWrapper.tsx
JennaXiao3: ahhhh okie thanks!!!

frontend/src/components/assessments/assessment-creation/StudentDashboardHeader.tsx
JennaXiao3: yep they are removing it

joyce-shi (approved):
looks awesome! a bit cleaner now 😎

PR #237, opened 2023-05-02
feat: add dynamic filter options

carissa-tang:
## Notion ticket link

[Dynamically populate filter menu](https://www.notion.so/uwblueprintexecs/Dynamically-populate-filter-menu-09554745745946ec8114b84861ae8044?pvs=4)

## Implementation description
* Use result of `GET_ALL_TESTS` to populate filter options
* Update `titleCase` to handle cases like ` a` or `a    a`

## Steps to test
1. Go to http://localhost:3000/admin/assessments and use filter

* Should we sort the filter options? (i.e. lexicographically)

## Checklist


frontend/src/constants/CreateAssessmentConstants.ts
```diff
@@ -1,9 +1,51 @@
...
 
-export default gradeOptions;
+export interface TestFilterOptions {
+  gradeOptions: StringOption[];
+  testTypeOptions: StringOption[];
+  countryOptions: StringOption[];
+  regionOptions: StringOption[];
+}
+
+export const getTestFilterOptions = (
```
joyce-shi: could we move this function from `constants` to `utils`? may need to create the appropriate file. honestly could probably also move `gradeOptions` there too (& then get rid of the file?)

also thoughts on renaming to `testFilterOptions` to be consistent with `gradeOptions`? (or can change to `getGradeOptions`, just wanna be consistent!

alsoooooo we should probably be consistent about either using `test` or `assessment`. i think for now, we've been using `assessment` on the frontend and `test` on the backend, but the lines are starting to blur haha

frontend/src/constants/CreateAssessmentConstants.ts
```diff
@@ -1,9 +1,51 @@
...
+      label: titleCase(value),
+    })),
+    countryOptions: Array.from(countriesSet).map((value) => ({
+      value,
+      label: titleCase(value),
+    })),
+    regionOptions: Array.from(regionsSet).map((value) => ({
+      value,
+      label: titleCase(value),
+    })),
```
joyce-shi: thoughts on refactoring the `Array.from...` logic to a function for a cleaner implementation?

joyce-shi (commented):
overall looks good, left a couple comments!

also, yes i think it makes so much sense to sort filer options! was just gonna comment on how it loooks kind of off w/o it haha

frontend/src/components/pages/admin/DisplayAssessmentsPage.tsx
```diff
@@ -55,112 +43,62 @@ const DisplayAssessmentsPage = (): React.ReactElement => {
...
   const searchedAssessements = React.useMemo(() => {
-    let filteredTests = filteredAssessements;
-    if (search) {
-      filteredTests = filteredTests.filter(
-        (assessment: AssessmentProperties) =>
-          assessment.name.toLowerCase().includes(search.toLowerCase()) ||
-          removeUnderscore(assessment.grade)
-            .toLowerCase()
-            .includes(search.toLowerCase()) ||
-          getFirstNumber(assessment.grade)
```
carissa-tang, line 126: i'm not sure why this was needed in the first place?? so i removed it. but lmk if we need it

frontend/src/utils/GeneralUtils.ts
```diff
@@ -12,9 +11,8 @@ export const removeUnderscore = (word: string): string => {
   return word.replace(/_/g, " ");
 };
 
-export const getFirstNumber = (word: string): string => {
-  const numbers = word.match(/^\d+/);
-  return numbers ? numbers[0] : "";
+export const includesIgnoreCase = (text: string, pattern: string): boolean => {
```
carissa-tang, line 14: think of a better name

joyce-shi (approved):
LGTM

PR #274, opened 2023-05-11
feat: Handle Unverified Users Trying to Log In

carissa-tang:
## Notion ticket link

[Handle Unverified Users Trying to Log In](https://www.notion.so/uwblueprintexecs/Handle-Unverified-Users-Trying-to-Log-In-223c5520a124475d893b5a8607a26ff7?pvs=4)

## Implementation description
* When unverified teacher tries to login, they are redirected to a new "Verify Your Email" screen
* Note: admins cannot login without being verified - they are auto verified when they set their password

**Changes**
* Return `emailVerified` field in `login` resolver
   * True if email is verified
* Only set cookies and localStorage values if user is verified  
* Add `sendEmailVerificationLink` resolver / mutation
* `UnverifiedUsers` component -> shown if unverified user tries to login

## Steps to test
1. Create a new teacher account - don't verify it
2. Try to log in

Note: if you enter the wrong password, the existing window alert pops up - should we handle this differently, as might be a security hazard?

https://github.com/uwblueprint/jump-math/assets/69697180/41e401eb-c18d-4328-9845-6a5b45b1ff3f

* UI matches Figma
* No breaking changes

## Checklist


frontend/src/components/auth/Login.tsx
```diff
@@ -9,12 +9,13 @@ import { ADMIN_SIGNUP_IMAGE, TEACHER_SIGNUP_IMAGE } from "../../assets/images";
 import * as Routes from "../../constants/Routes";
 import { HOME_PAGE } from "../../constants/Routes";
 import AuthContext from "../../contexts/AuthContext";
-import { AuthenticatedUser } from "../../types/AuthTypes";
+import { AuthenticatedUser, VerifiableUser } from "../../types/AuthTypes";
```
carissa-tang, line 12:
```suggestion
import { VerifiableUser } from "../../types/AuthTypes";
```

frontend/src/components/auth/reset-password/ForgotPassword.tsx
```diff
@@ -1,22 +1,16 @@
...
 
 import { RESET_PASSWORD } from "../../../APIClients/mutations/AuthMutations";
 import { GET_USERS_BY_ROLE } from "../../../APIClients/queries/UserQueries";
 import { UserResponse } from "../../../APIClients/types/UserClientTypes";
-import { ArrowBackOutlineIcon } from "../../../assets/icons";
 import {
   ADMIN_SIGNUP_IMAGE,
   TEACHER_SIGNUP_IMAGE,
 } from "../../../assets/images";
+import LoadingState from "../../common/LoadingState";
```
carissa-tang, line 12:
```suggestion
```

frontend/src/components/auth/UnverifiedUsers.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [sendEmailVerificationLink] = useMutation<{ resetPassword: boolean }>(
+    SEND_EMAIL_VERIFICATION_LINK,
+  );
+
+  return (
+    <AuthWrapper
+      form={
+        <ResendEmail
+          resendFunction={async () => {
+            await sendEmailVerificationLink({ variables: { email } });
```
jfdoming, line 20: nit: probably don't need async-await here, can just return the promise directly

frontend/src/components/common/ResendEmail.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <>
+      <Text textAlign="center" textStyle="subtitle2">
+        Didn’t receive the email?{" "}
+        <Button
+          color="blue.300"
+          display="contents"
+          onClick={onResendClick}
+          style={{ font: "inherit" }}
```
jfdoming: just a thought, maybe we could consider disabling the button/putting it in a loading state (or just making the click do nothing) while the mutation is running? just so people don't accidentally send multiple emails to themselves

jfdoming (approved):
Out of curiosity, why was the extra `VerifiableUser` type needed? LGTM though

frontend/src/components/auth/UnverifiedUsers.tsx
carissa-tang: there's a type-mismatch btwn `Promise<FetchResult<{ resetPassword: boolean; }>>` and `() => Promise<void>`. i figured keeping a more generic type and using the wrapper would be cleaner

frontend/src/components/common/ResendEmail.tsx
carissa-tang: good idea, thanks!

carissa-tang:
> VerifiableUser

I didn't directly add the field to `AuthenticatedUser` or `AuthenticatedAdminOrTeacher` if that's your question bc it gets complicated with `AuthenticatedUser` being in local storage but we can't store booleans (also there's no need to store it). i think it also got messy with the union being nullable

PR #276, opened 2023-05-13
backend: Archive Classroom

cindy-531:
## Notion ticket link

[Update Backend to Archive Classrooms](https://www.notion.so/uwblueprintexecs/Update-backend-to-archive-classrooms-5e9d21e6f91c415388c1d0e7db24aeaa?pvs=4)

## Implementation description
* Update the backend to support the Archive Classroom feature

## Steps to test
1. Check that yarn test runs with all test cases passed

## Checklist


backend/models/class.model.ts
```diff
@@ -21,6 +21,8 @@ export interface Class extends Document {
   testSessions: string[];
   /** the students of the class */
   students: Student[];
+  /** the status of the class */
```
joyce-shi: nit: 
```suggestion
  /** whether the class is active or archived */
```

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -111,7 +113,7 @@ describe("mongo classService", (): void => {
   it("getClassById for invalid Id", async () => {
     // execute and assert
     const notFoundId = "56cb91bdc3464f14678934cd";
-    await ClassModel.create(testClass[0]);
+    await ClassModel.create(mockTestClass);
```
joyce-shi: can we update the cases in `testClass` instead of creating a new `mockTestClass`? i think all the test cases should probably be updated to include the `isActive` field anyway, yeah?

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -174,8 +176,19 @@ describe("mongo classService", (): void => {
...
+  it("archive class", async () => {
+    // add test class
+    const classObj = await ClassModel.create(mockTestClass);
+
+    // execute
+    const res = await classService.archiveClass(classObj.id);
+
+    // assert
+    expect(res.isActive).toEqual(false);
+  });
```
joyce-shi, line 205: can we add 2 test cases? 
1. when the id is not found 
2. when the class is already archived

backend/services/implementations/classService.ts
```diff
@@ -191,6 +193,32 @@ class ClassService implements IClassService {
     }
   }
 
+  async archiveClass(id: string): Promise<ClassResponseDTO> {
+    let archivedClass: Class | null;
+    try {
+      archivedClass = await MgClass.findOneAndUpdate(
+        { _id: id },
```
joyce-shi: do we need to add this? i think we should probably throw an error if the class is already archived
```suggestion
        { _id: id, isActive: true },
```

backend/services/implementations/classService.ts
```diff
@@ -191,6 +193,32 @@ class ClassService implements IClassService {
...
+        { isActive: false },
```
joyce-shi: i think we have to use the `$set` operator here? 

see more under `update` parameter: https://www.mongodb.com/docs/manual/reference/method/db.collection.findOneAndUpdate/

Screenshot 2023-05-29 at 11 52 49 AM

let me know if i'm misinterpreting this!

backend/services/implementations/classService.ts
```diff
@@ -191,6 +193,32 @@ class ClassService implements IClassService {
...
+        {
+          new: true,
+          runValidators: true,
+        },
+      );
+
+      if (!archivedClass) {
+        throw new Error(`Class with id ${id} not found`);
```
joyce-shi:
```suggestion
        throw new Error(`Class with id ${id} not found or not currently active`);
```

joyce-shi (commented):
awesome! looks good, left a couple comments. (: 

in the description of this PR, could you also link a test of your new resolver in the graphql playground? (i.e screenshot / video). so that whoever adds the frontend feature can reference it!

backend/services/implementations/classService.ts
cindy-531: If it's not currently active, we have to explicitly check for it

backend/graphql/index.ts
```diff
@@ -96,7 +96,7 @@ const buildSchema = async () => {
       duplicateTest: authorizedByAdmin(),
       unarchiveTest: authorizedByAdmin(),
       archiveTest: authorizedByAdmin(),
-      createClass: authorizedByAllRoles(),
+      // createClass: authorizedByAllRoles(),
```
carissa-tang: can you undo this?

backend/services/interfaces/classService.ts
```diff
@@ -90,6 +91,14 @@ export interface IClassService {
    */
   deleteClass(id: string): Promise<string>;
 
+  /**
+   * This method archives the class the given id
```
carissa-tang:
```suggestion
   * This method archives the class with the given id
```

backend/services/implementations/classService.ts
```diff
@@ -61,6 +61,7 @@ class ClassService implements IClassService {
       className: newClass.className,
       startDate: newClass.startDate,
       gradeLevel: newClass.gradeLevel,
+      isActive: newClass.isActive,
```
carissa-tang, line 64: I think you will need to update line 50 to have something like 
```
newClass = await MgClass.create({
        ...classObj,
        isActive: true,
        testSessions: [],
        students: [],
      });
```
assuming that classes are always active when first created

carissa-tang (changes requested):

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -174,8 +176,36 @@ describe("mongo classService", (): void => {
...
+  it("archive class with non-existing id", async () => {
+    const notFoundId = "86cb91bdc3464f14678934cd";
+    await expect(async () => {
+      await classService.archiveClass(notFoundId);
+    }).rejects.toThrowError(`Class with id ${notFoundId} not found`);
+  });
+
+  it("archive class that is already archived", async () => {
+    const classObj = await ClassModel.create(mockTestClass);
+    const archivedClass = await classService.archiveClass(classObj.id);
```
carissa-tang: rather than making the call twice, i would prefer if you made a new constant or smth like
```suggestion
    const classObj = await ClassModel.create({...mockTestClass, isActive: false});
```
just a best practice, to isolate what we are testing

backend/services/implementations/__tests__/classService.test.ts
carissa-tang: `testClass` is the request object, which won't contain that field. so should be fine to leave like this i think

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -63,10 +65,10 @@ describe("mongo classService", (): void => {
 
   it("create class for valid teachers", async () => {
     // execute
-    const createdClass = await classService.createClass(testClass[0]);
+    const createdClass = await classService.createClass(mockTestClass);
 
     // assert
-    assertResponseMatchesExpected(testClass[0], createdClass);
+    assertResponseMatchesExpected(mockTestClass, createdClass);
```
carissa-tang: can you update `assertResponseMatchesExpected` to check for the `isActive` field?

carissa-tang (changes requested):

carissa-tang (approved):

PR #285, opened 2023-05-14
feat: test sessions page

jfdoming:
## Notion ticket link

[Test Sessions page](https://www.notion.so/uwblueprintexecs/Test-Sessions-page-da4a9b79d4c44a7781088909135d14d8)

## Implementation description
image

## Steps to test
1. Comment out parts of the code/change the error/loading states to see the different UI states.

* 

## Checklist


frontend/src/components/assessments/TestSessionListItem.tsx
```diff
@@ -18,24 +18,27 @@ import Copyable from "../common/Copyable";
...
 
+export const STATUSES = ["active", "upcoming", "past"] as const;
+export type TestSessionStatus = typeof STATUSES[number];
+
+type TestSessionItemStats = {
+  mean: number;
+  median: number;
+  completionRate: number;
+  submissions: number;
+};
```
joyce-shi: thoughts on moving `TestSessionStatus` and `TestSessionStats` to  types file? (you would have to create a new one). rationale: b/c these are shared types among some other features and files on the Teacher Dashboard as well.

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+          ]}
+          subtitle="You currently have no assessments."
+          textColor="blue.300"
+        />
+      )}
+    </>
+  );
 };
 
 export default DistributeAssessmentPage;
```
joyce-shi: could we actually name this something like `TestSessionsPage` or `DisplayTestSessionsPage`? `DistributeAssessmentPage` is a different flow (what you've linked as `CREATE_ASSESSMENT_PATH`). you'll need to create a new route for your page as well, which you can put under `TeacherRouting`.

sorry, i think i should have been more clear in the ticket! (i thought a `TestSessionsPage` already existed, but I don't think it does)

frontend/src/components/assessments/TestSessionListItem.tsx
```diff
@@ -18,24 +18,27 @@ import Copyable from "../common/Copyable";
 import Popover from "../common/Popover";
 import PopoverButton from "../common/PopoverButton";
 
+export const STATUSES = ["active", "upcoming", "past"] as const;
+export type TestSessionStatus = typeof STATUSES[number];
```
joyce-shi: out of curiosity, do we need to separate this into two lines? is this the same thing as `type TestSessionStatus = "active" | "upcoming" | "past"`?

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+        <HStack justifyContent="space-between">
+          <Text
+            color="blue.300"
+            marginBottom="0.5em"
+            style={{ textAlign: "left" }}
+            textStyle="header4"
+          >
+            Assessments
+          </Text>
+          {!!data?.length && !loading && !error && (
```
joyce-shi: I think the idea is to show this button even when the data has not loaded. that should be consistent w/ other table implementations!

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+            <Button
+              mt={10}
+              onClick={() => history.push(CREATE_ASSESSMENT_PATH)}
+              rightIcon={<PlusOutlineIcon />}
+              variant="primary"
+            >
+              Add Assessment
+            </Button>
```
joyce-shi: nit: I think a lot of pages in our app use this header + button pattern (like 4 or 5?). what are thoughts on refactoring it to a common component and then refactoring the existing components to use it? :D (could be done in a subsequent PR)

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+    error: false,
+  };
+
+  const dataWithStatus = useMemo(() => {
+    const now = new Date();
+    return data.map((session) => {
+      const status: TestSessionStatus = (() => {
+        if (session.endDate < now) return "past";
+        if (session.startDate > now) return "upcoming";
+        return "active";
```
joyce-shi: thoughts on refactoring this logic into a function for readability? & maybe moving it into a utils folder? I think we'll probably use the logic very soon in places coming up. (but not that many, so i'm fine leaving it where it is as well)

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+        <MessageContainer
+          buttonIcon={<PlusOutlineIcon />}
+          buttonRoute={CREATE_ASSESSMENT_PATH}
+          buttonText="Create new assessment"
+          image={DistributeAssessmentsIllustration}
+          paragraphs={[
+            "Click on the button below to create your first assessment.",
+          ]}
+          subtitle="You currently have no assessments."
+          textColor="blue.300"
```
joyce-shi: nit: thoughts on refactoring the `MessageContainer` to a separate component? i think that would be consistent with what we've done in the past. and also for readability & b/c this page is getting a bit chunky. but i'm fine leaving it as is as well (up to you!)

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+    return filteredData.sort((a, b) => {
+      return (
+        (invertSort ? -1 : 1) *
+        (a.targetDate.getTime() - b.targetDate.getTime())
+      );
+    });
+  }, [filteredData, currentTab]);
+
+  const pageSize = 8;
+  const numPages = Math.ceil(sortedData.length / pageSize);
```
joyce-shi: what are thoughts on moving this to a separate file, i.e `PaginationUtils`? & then refactoring some of the calculations in `Table` to use this as well? so we can keep the page size between the two "table" implementations consistent & making some of these pagination calculations more readable?

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+                <Tab key={status} onClick={() => setCurrentTab(status)}>
+                  {titleCase(status)}
+                </Tab>
+              ))}
+            </TabList>
+            <TabPanels>
+              {STATUSES.map((status) => (
+                <TabPanel key={status}>
+                  {sortedData
+                    .slice((currentPage - 1) * 8, currentPage * 8)
```
joyce-shi: -> variables for readability?

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -1,8 +1,247 @@
...
+    completionRate: number;
+    submissions: number;
+  };
+  startDate: Date;
+  endDate: Date;
+  testName: string;
+}[] = [
+  ...[...Array(20)].map((_, i) => ({
+    id: i.toString(),
+    accessCode: "086 731",
```
joyce-shi: if you fetch the access codes, i think they probably don't have a space in the middle & design added a subsequent space for readability. so you would have to add the space on the fe

joyce-shi (commented):
overall looks good to me! just left some minor thoughts re: readability & style. not married to any of them, so feel free to disagree! 😄 

as a side note: we've had a hard time keeping some of these table pages clean and slim, so sorry for all the nit picks. it's been a task in the backlog to clean some of those pages up for a loooong while, so that's always a wip on this project 😓

frontend/src/components/assessments/TestSessionListItem.tsx
jfdoming: Not quite, I wanted to have an array of statuses I could use in JS. The first line creates a JS array, and the second creates a type alias which is a union of all the array elements. Unfortunately you need both in order to use the type in both TS and JS.

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
jfdoming: Yes good idea! Subsequent PR is the way to go though

frontend/src/components/assessments/TestSessionListItem.tsx
joyce-shi: gotcha

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
jfdoming: Agree. Let's do it in a follow-up PR though

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
jfdoming: sure

joyce-shi (approved):
🚢

PR #294, opened 2023-05-15
feat: fetch data for test sessions page

jfdoming:
## Notion ticket link

[Fetch data for Test Sessions page](https://www.notion.so/uwblueprintexecs/Fetch-data-for-Test-Sessions-page-29a10c0026654c0ab2ba0a19db62e86f)

Strongly suggest reviewing by commit!

- Fetch data from backend
- Fetch className from backend
- Fix linter script on Mac OS

image

## Implementation description
* As discussed on Slack, decided to add a reference to the classroom into the session object. Alternatively would require multiple fetches.

## Steps to test
1. View the teacher session page and verify that it loads and you see live data (probably nothing :P).

* 

## Checklist

### DIFF
--- a/backend/services/implementations/__tests__/classService.test.ts
+++ b/backend/services/implementations/__tests__/classService.test.ts
@@ -5,22 +5,28 @@ import db from "../../../testUtils/testDb";
 import {
   testClass,
   testClassInvalidTeacher,
-  assertResponseMatchesExpected,
   updatedTestClass,
   testStudents,
   testClassWithStudents,
   updatedTestClassWithStudent,
   updatedTestStudents,
-  assertStudentResponseMatchesExpected,
-  testClassWithTestSessions,
-  assertArrayResponseMatchesExpected,
 } from "../../../testUtils/class";
+import {
+  assertResponseMatchesExpected,
+  assertArrayResponseMatchesExpected,
+  assertStudentResponseMatchesExpected,
+} from "../../../testUtils/classAssertions";
 import UserService from "../userService";
 import { mockTeacher } from "../../../testUtils/users";
 import TestSessionService from "../testSessionService";
 import TestService from "../testService";
 import SchoolService from "../schoolService";
 import { mockTestSessionWithId } from "../../../testUtils/testSession";
+
+const testClassWithTestSessions = {
+  ...testClass[0],
+  testSessions: [mockTestSessionWithId.id],
+};
 
 describe("mongo classService", (): void => {
   let classService: ClassService;
--- a/backend/services/implementations/classService.ts
+++ b/backend/services/implementations/classService.ts
@@ -9,7 +9,10 @@ import {
   StudentRequestDTO,
 } from "../interfaces/classService";
 import IUserService from "../interfaces/userService";
-import { ITestSessionService } from "../interfaces/testSessionService";
+import {
+  ITestSessionService,
+  TestSessionResponseDTO,
+} from "../interfaces/testSessionService";
 
 const Logger = logger(__filename);
 
@@ -58,7 +61,10 @@ class ClassService implements IClassService {
     };
   }
 
-  async getClassById(id: string): Promise<ClassResponseDTO> {
+  async getClassById(
+    id: string,
+    populateTestSessions = true,
+  ): Promise<ClassResponseDTO> {
     let classObj: Class | null;
     try {
       classObj = await MgClass.findById(id);
@@ -69,7 +75,9 @@ class ClassService implements IClassService {
       Logger.error(`Failed to get Class. Reason = ${getErrorMessage(error)}`);
       throw error;
     }
-    return (await this.mapClassToClassDTOs([classObj]))[0];
+    return (
+      await this.mapClassToClassDTOs([classObj], populateTestSessions)
+    )[0];
   }
 
   async getClassByTestSessionId(
@@ -113,6 +121,7 @@ class ClassService implements IClassService {
 
   private async mapClassToClassDTOs(
     classObjs: Array<Class>,
+    populateTestSessions = true,
   ): Promise<Array<ClassResponseDTO>> {
     const classDtos: Array<ClassResponseDTO> = await Promise.all(
       classObjs.map(async (classObj) => {
@@ -120,11 +129,14 @@ class ClassService implements IClassService {
           classObj.teacher,
         );
 
-        const testSessionIds = classObj.testSessions;
-        const testSessionPromises = testSessionIds.map((id) =>
-          this.testSessionService.getTestSessionById(id),
-        );
-        const testSessionDTOs = await Promise.all(testSessionPromises);
+        let testSessionDTOs: Array<TestSessionResponseDTO> = [];
+        if (populateTestSessions) {
+          const testSessionIds = classObj.testSessions;
+          const testSessionPromises = testSessionIds.map((id) =>
+            this.testSessionService.getTestSessionById(id),
+          );
+          testSessionDTOs = await Promise.all(testSessionPromises);
+        }
 
         return {
           id: classObj.id,
--- a/backend/services/interfaces/classService.ts
+++ b/backend/services/interfaces/classService.ts
@@ -1,4 +1,4 @@
-import { type UserDTO, type Grade } from "../../types";
+import type { UserDTO, Grade } from "../../types";
 import { type TestSessionResponseDTO } from "./testSessionService";
 
 export interface StudentRequestDTO {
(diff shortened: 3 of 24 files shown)


scripts/lint.sh
```diff
@@ -4,7 +4,13 @@ set -e
 
 # https://stackoverflow.com/a/32981392/3761440
 faketty () {
-    script -qefc "$(printf "%q " "$@")" /dev/null
```
jfdoming, line 7: macos `script` command doesn't have an `-e` option

scripts/lint.sh
```diff
@@ -4,7 +4,13 @@ set -e
 
 # https://stackoverflow.com/a/32981392/3761440
```
carissa-tang: maybe update this link if there's a more relevant one? or just remove

frontend/src/APIClients/types/TestSessionClientTypes.ts
```diff
@@ -1,3 +1,5 @@
+import { Test } from "./TestClientTypes";
+
```
carissa-tang, line 3: just noticed that `endDate` is not actually needed in places where `TestSessionMetadata` is used i think and we can make a small refactor? correct me if i'm wrong tho

frontend/src/APIClients/types/TestSessionClientTypes.ts
```diff
@@ -8,3 +10,10 @@ export interface TestSessionMetadata {
   /** notes inputted by teacher to show students prior to commencing the test */
   notes?: string;
 }
+
```
carissa-tang, line 12: the names are a bit confusing here - would it make sense to name them after their purpose since their use cases are pretty specific? smth like `PreAssessmentMetadata` and `DuringAssessmentMetadata` or `DisplayAssessmentsMetadata`. what is best practice?

frontend/src/components/common/table/usePaginatedData.ts
```diff
@@ -1,29 +1,32 @@
 import { useMemo, useState } from "react";
 
-type PaginatedDataResult<T> = {
-  paginatedData: T[];
+type PaginatedDataResult<E, Data extends Array<E> | undefined> = {
```
carissa-tang: what does `E` here mean? element? lol

backend/services/interfaces/classService.ts
```diff
@@ -1,5 +1,5 @@
-import { UserDTO, Grade } from "../../types";
-import { TestSessionResponseDTO } from "./testSessionService";
+import { type UserDTO, type Grade } from "../../types";
+import { type TestSessionResponseDTO } from "./testSessionService";
```
carissa-tang, line 2: @joyce-shi maybe we make a ticket (in backlog) to add type-only imports to rest of codebase?

backend/services/interfaces/classService.ts
```diff
@@ -1,5 +1,5 @@
-import { UserDTO, Grade } from "../../types";
-import { TestSessionResponseDTO } from "./testSessionService";
+import { type UserDTO, type Grade } from "../../types";
```
carissa-tang:
```suggestion
import type { UserDTO, Grade } from "../../types";
```

frontend/src/utils/TestSessionUtils.ts
```diff
@@ -1,21 +1,21 @@
 import { TestSessionStatus } from "../types/TestSessionTypes";
 
 export const getSessionStatus = (
-  startDate: Date,
-  endDate: Date,
+  startDate: string | Date,
```
carissa-tang, line 4: wondering what's the need for string?

carissa-tang (commented):
got through most of it, but will take a second look tmr!!

frontend/src/APIClients/types/TestSessionClientTypes.ts
jfdoming: Yeah I don't think it's used, which is odd though... I feel like knowing the end date is useful for students, but anyways. Will refactor if we think it won't ever be used (I could see a use-case where we auto-submit the test after the end-date passes though or something) cc @joyce-shi

frontend/src/components/common/table/usePaginatedData.ts
jfdoming: Yeahhhhh could be more clear tbf, will change to `Entry`

frontend/src/utils/TestSessionUtils.ts
jfdoming: The date actually comes from the backend as a string, not a `Date`, so with the current use it's always a string. I figured there might be a situation in the future where we already had a `Date` though, hence the union.

frontend/src/APIClients/types/TestSessionClientTypes.ts
jfdoming: I'm thinking something like `TestSessionMetadata` -> `TestSessionSetupData` and `TestSession` -> `TestSessionOverviewData`

backend/services/interfaces/classService.ts
joyce-shi: good idea! I think this can be fixed by adding the consistent-type-imports rule to the linter & running lint —fix, so should be quick. I’ll add a ticket to this sprint for this and assign to myself.

frontend/src/APIClients/types/TestSessionClientTypes.ts
joyce-shi: oops this is definitely an oversight on my end. If not used, feel free to remove!

backend/services/implementations/__tests__/classService.test.ts
```diff
@@ -5,23 +5,29 @@ import db from "../../../testUtils/testDb";
...
+  assertStudentResponseMatchesExpected,
+} from "../../../testUtils/classAssertions";
 import UserService from "../userService";
 import { mockTeacher } from "../../../testUtils/users";
 import TestSessionService from "../testSessionService";
 import TestService from "../testService";
 import SchoolService from "../schoolService";
 import { mockTestSessionWithId } from "../../../testUtils/testSession";
 
+const testClassWithTestSessions = {
```
carissa-tang, line 26: what's the reason for moving this?

frontend/src/APIClients/types/TestSessionClientTypes.ts
```diff
@@ -1,10 +1,24 @@
-export interface TestSessionMetadata {
+import { ClassResponse } from "./ClassClientTypes";
+import { Test } from "./TestClientTypes";
+
+interface TestSessionMetadata {
```
carissa-tang, line 4: what's the reason for having `TestSessionMetadata` and not just `TestSessionSetupData` directly?

backend/services/implementations/classService.ts
```diff
@@ -113,18 +121,22 @@ class ClassService implements IClassService {
 
   private async mapClassToClassDTOs(
     classObjs: Array<Class>,
+    populateTestSessions = true,
```
carissa-tang, line 124: is there infinite loop without this? was wondering if there's a more elegant way of handling, but maybe not

frontend/src/utils/TestSessionUtils.ts
carissa-tang, line 4: should we be using `Date` throughout for consistency? and handle conversions in apollo https://www.apollographql.com/docs/apollo-server/schema/custom-scalars/#example-the-date-scalar

carissa-tang (approved):
i see you enabled auto-merge 👀 just left a few thoughts out of curiosity !

backend/services/implementations/__tests__/classService.test.ts
jfdoming: Dependency cycle... Wish I didn't have to but this was the easiest way

frontend/src/APIClients/types/TestSessionClientTypes.ts
jfdoming: My thinking was that `TestSessionMetadata` would contain the common fields between the two other types. Currently they're equal but conceptually my thinking was that they could be different.

backend/services/implementations/classService.ts
jfdoming: Yeah, there is! I suppose we could also create a second class method instead to handle the no-hydration logic, but this seemed like the simplest way to go. Open to suggestions though in a follow-up PR

frontend/src/utils/TestSessionUtils.ts
jfdoming: Agree! I will make a PR for this

PR #315, opened 2023-05-17
chore: adding consistent-imports rule to linter

joyce-shi:
## Notion ticket link

[Fix type imports](https://www.notion.so/uwblueprintexecs/Fix-type-imports-3533a042f5014e53a193b3fdafd71e9b?pvs=4)

## Implementation description
* added `consistent-imports` rule to linter as discussed in #294  
* and ran linter fix

* no breaking changes

### DIFF
--- a/backend/graphql/resolvers/entityResolvers.ts
+++ b/backend/graphql/resolvers/entityResolvers.ts
@@ -1,7 +1,7 @@
 import fs from "fs";
-import { FileUpload } from "graphql-upload";
+import type { FileUpload } from "graphql-upload";
 /* eslint-disable-next-line import/no-extraneous-dependencies */
-import { ReadStream } from "fs-capacitor";
+import type { ReadStream } from "fs-capacitor";
 import multer from "multer";
 import {
   validateFileType,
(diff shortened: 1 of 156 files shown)


backend/graphql/resolvers/entityResolvers.ts
```diff
@@ -1,15 +1,15 @@
 import fs from "fs";
-import { FileUpload } from "graphql-upload";
+import type { FileUpload } from "graphql-upload";
 /* eslint-disable-next-line import/no-extraneous-dependencies */
```
jfdoming: Might be able to remove this line now?

backend/services/implementations/imageUploadService.ts
```diff
@@ -1,18 +1,18 @@
 /* eslint-disable-next-line import/no-extraneous-dependencies */
```
jfdoming: Ditto, maybe?

jfdoming (approved):

backend/graphql/resolvers/entityResolvers.ts
joyce-shi: yes, good catch!

PR #320, opened 2023-05-18
fix: test session teacher page style fixes

jfdoming:
## Notion ticket link

[Test session teacher page style fixes](https://www.notion.so/uwblueprintexecs/Test-session-teacher-page-style-fixes-138c50f42942427688d77794cb9f9d61)
* Change classroom-tag width to fit content better
* Make past sessions clickable

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/frontend/src/components/sessions/TestSessionListItem.tsx
+++ b/frontend/src/components/sessions/TestSessionListItem.tsx
@@ -83,7 +83,7 @@ const TestSessionListItem = ({
           <Tag
             bg="blue.50"
             borderRadius="full"
-            maxWidth={36}
+            maxWidth={40}
             overflow="hidden"
             size="lg"
           >
--- a/frontend/src/components/sessions/TestSessionListItem.tsx
+++ b/frontend/src/components/sessions/TestSessionListItem.tsx
@@ -67,9 +67,9 @@ const TestSessionListItem = ({
         <Tag
           bg="blue.50"
           borderRadius="full"
+          maxWidth={36}
           overflow="hidden"
           size="lg"
-          w={36}
         >
           <TagLeftIcon aria-hidden="true" as={BookIcon} color="blue.300" />
           <TagLabel>
--- a/frontend/src/components/sessions/TestSessionListItem.tsx
+++ b/frontend/src/components/sessions/TestSessionListItem.tsx
@@ -55,77 +55,94 @@ const TestSessionListItem = ({
   )} ${accessCode.slice(ACCESS_CODE_GROUP_SIZE)}`;
 
   return (
-    <HStack gap={2} pb={8} w="100%">
-      <Tooltip
-        bg="blue.300"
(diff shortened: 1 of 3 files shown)


frontend/src/components/sessions/TestSessionListItem.tsx
```diff
@@ -55,77 +55,94 @@ const TestSessionListItem = ({
...
-          </Text>
-          <Text color="grey.300" textStyle="mobileParagraph">
-            Median: {stats.median}%
-          </Text>
-          <Text color="grey.300" textStyle="mobileParagraph">
-            Completion Rate: {stats.completionRate}%
+          <Tag
+            bg="blue.50"
+            borderRadius="full"
+            maxWidth={36}
```
joyce-shi:
```suggestion
            maxWidth={40}
```
To match design spec of 161px width? Let me know if i'm converting that incorrectly

joyce-shi (approved):
lgtm! a couple side notes on this page!
- the bottom of the page might benefit from some padding? looks like the pagination table number are flush against the bottom of the screen. not sure if it was actually my last PR that caused this (if so, sorry!)
- i think that past assessments are not supposed to be editable (i.e the popover should not have edit option)
- i feel like the `TestSessionListItem` component might be getting bit distracted...? since its covering a lot of different cases now. I wonder if it makes sense to refactor it into a parent component with children and then having separate components for `Active`, `Upcoming` and `Past` components? but not super high priority, so would be totally fine if we didn't refactor or left it in the backlog (which probably make more sense than doing it now).

frontend/src/components/sessions/TestSessionListItem.tsx
jfdoming: Yes good suggestion! For the other points above, I addressed the first one. For the second one, I'll address this in the other PR since otherwise there will be merge conflicts. For the last one, I think it would be quite difficult to make separate components without duplicating most of the layout and functionallty. Perhaps turning some of these style props into additional map-objects rather than including ternary statements might help? Let me know what you think is best, but I'm in favour of avoiding a refactor unless the components really differ a lot functionally by status.

jfdoming:
[Follow-up ticket](https://www.notion.so/uwblueprintexecs/47bf4727a2444bccad4a91952898b8f3?v=9658436bf9be409e90581631a7155f16&p=ed52aeb4fa4147a79e47baaadfa80e58&pm=s) to fix the non-urgent button-nesting issue I mentioned on Slack. Will probably delay this until after I make a dent in some of the other tickets I'm assigned, since the current implementation works well, it's just _technically_ invalid HTML 😬

frontend/src/components/sessions/TestSessionListItem.tsx
joyce-shi: okay sounds good. no refactor 👍 i'm happy with as is 😀

PR #327, opened 2023-05-21
backend: Compute Test Completion Rate

wais-hundekar:
## Notion ticket link

[Compute Test Completion Rate](https://www.notion.so/uwblueprintexecs/47bf4727a2444bccad4a91952898b8f3?v=9658436bf9be409e90581631a7155f16&p=a43af723a4d74def91ada778cb9baa22&pm=s)

## Implementation description
* Added getCompletionRateByTest and appropriate utility functions

## Steps to test
1. Ensure existing tests are accurate

* 

## Checklist


backend/utilities/generalUtils.ts
```diff
@@ -19,3 +19,11 @@ export const calculateMedianScore = (values: number[]): number => {
   }
   return (values[mid] + values[mid - 1]) / 2;
 };
+
+export const isCompletedTestSession = (
```
carissa-tang: can we rename to `isCompletedTestResult`?

backend/utilities/generalUtils.ts
```diff
@@ -19,3 +19,11 @@ export const calculateMedianScore = (values: number[]): number => {
...
+  results: Array<Array<Array<number>>>,
+): boolean => {
+  return results.every((question) =>
+    question.every((subquestion) => subquestion.length !== 0),
+  );
```
carissa-tang:
```suggestion
  results.flat().every((answer) => answer.length);
```
might be a little easier to read?

carissa-tang (commented):
just a few nits!

jfdoming (commented):
Broadly speaking, I was also wondering about performance. I saw at one point we had been thinking about memoizing/incrementally computing these statistics. Did we ever come to a decision about whether to do that? It seems to me that computing these every time we want to retrieve them might be slower, and it seems straightforward to maintain a `numCompletedResults` count and update it when a test is completed. That said, I don't have all the context here so I don't want to go against a decision which was already made.

Alternatively, we could even just develop a more generalizable method of caching these statistics based on a `dirty` flag we add to the `TestSession` model. (That way we can reuse it for other statistics.) This has the advantage that the logic in other endpoints is much simpler. I also prefer this method because it maintains separation-of-concerns, but up to you if this is something we care about.

wais-hundekar:
> Broadly speaking, I was also wondering about performance. I saw at one point we had been thinking about memoizing/incrementally computing these statistics. Did we ever come to a decision about whether to do that? It seems to me that computing these every time we want to retrieve them might be slower, and it seems straightforward to maintain a `numCompletedResults` count and update it when a test is completed. That said, I don't have all the context here so I don't want to go against a decision which was already made.
> 
> Alternatively, we could even just develop a more generalizable method of caching these statistics based on a `dirty` flag we add to the `TestSession` model. (That way we can reuse it for other statistics.) This has the advantage that the logic in other endpoints is much simpler. I also prefer this method because it maintains separation-of-concerns, but up to you if this is something we care about.

That sounds like a pretty interesting approach, I'd love to look into implementing it if its worthwhile, thoughts @carissa-tang @joyce-shi?

joyce-shi:
@jfdoming i could be wrong, but i think all 3 of us talked about this not too long ago? i thought the consensus was to leave it as brute force (since this set of statistics is done) & then to optimize if we have the bandwidth for it, but let me know if i'm misremembering.

jfdoming:
> @jfdoming i could be wrong, but i think all 3 of us talked about this not too long ago? i thought the consensus was to leave it as brute force (since this set of statistics is done) & then to optimize if we have the bandwidth for it, but let me know if i'm misremembering.

I don't remember the discussion, but I also have the memory of a gerbil who just downed a bottle of wine, so I agree let's push it to the backlog for now, and hopefully address it in the future. 🙌

jfdoming (approved):

PR #343, opened 2023-05-23
feat: add student list component

jfdoming:
## Notion ticket link

[Student list component](https://www.notion.so/uwblueprintexecs/Student-list-component-190349174c9947eab4e0290c1faa901e)

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/frontend/src/components/common/table/SortMenu.tsx
+++ b/frontend/src/components/common/table/SortMenu.tsx
@@ -18,12 +18,13 @@ import {
 import { FilterOptionsIcon } from "../../../assets/icons";
 import { titleCase } from "../../../utils/GeneralUtils";
 
+export type SortOrder = "ascending" | "descending";
 export interface SortMenuProps {
   properties: string[];
   labels: string[];
-  onSortProperty: React.Dispatch<React.SetStateAction<string>>;
-  onSortOrder: React.Dispatch<React.SetStateAction<string>>;
-  initialSortOrder?: string;
+  onSortProperty?: React.Dispatch<React.SetStateAction<string>>;
+  onSortOrder: React.Dispatch<React.SetStateAction<SortOrder>>;
+  initialSortOrder?: SortOrder;
 }
 
 const SortMenu = ({
@@ -34,9 +35,9 @@ const SortMenu = ({
   initialSortOrder = "ascending",
 }: SortMenuProps): React.ReactElement => {
   const [sortProperty, setSortProperty] = React.useState(properties[0]);
-  const [sortOrder, setSortOrder] = React.useState(initialSortOrder);
+  const [sortOrder, setSortOrder] = React.useState<SortOrder>(initialSortOrder);
 
-  const propertyList = (
+  const propertyList = properties.length ? (
     <RadioGroup
       color="blue.300"
       onChange={(e) => setSortProperty(e)}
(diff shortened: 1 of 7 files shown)


frontend/src/components/common/table/SortMenu.tsx
```diff
@@ -78,8 +79,8 @@ const SortMenu = ({
           <>
             <PopoverTrigger>
               <Button
-                leftIcon={<FilterOptionsIcon />}
-                minWidth="5%"
+                leftIcon={<SortOptionsIcon />}
```
jfdoming, line 82: For this change I noticed that the icon we're currently using doesn't match the one that appears in all of the designs. I updated the icon to reflect the designs, but please let me know if this decision was intentional and I can revert the change.

jfdoming:
~merging blocked by failing checks, which will be fixed by #345. PR is still reviewable, though!~ edit: resolved

frontend/src/components/common/table/SortMenu.tsx
joyce-shi: i'm fine with updating the icon! but is `FilterOptionsIcon` used elsewhere? if not, maybe we can remove from our icons page?

frontend/src/components/sessions/results/StudentList.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+type StudentListItemProps = {
+  firstName: string;
+  lastName: string;
+  isViewed: boolean;
+  isSelected: boolean;
+  onClick: () => void;
+};
+
+const StudentListItem = ({
```
joyce-shi, line 26: why do we keep this in the same component? vs. just separating out into its own?

joyce-shi (approved):
nice

frontend/src/components/sessions/results/StudentList.tsx
jfdoming: I didn't think the item component would be used outside this file, and I was hesitant to add an extra file since I can see the number of files in this particular directory getting quite large. If you have an idea for how I can organize the files such that making a separate file is clean, I'm more than happy to make the change!

frontend/src/components/common/table/SortMenu.tsx
jfdoming: It is not used elsewhere, but there are a lot of icons that aren't currently used on that page which we kept (I assume to make it easier to start components with those icons). If we don't already have a ticket, my preference would be to make a ticket to track removal of ALL the unused icons at the end of the project. Thoughts?

joyce-shi: makes sense to me!

frontend/src/components/sessions/results/StudentList.tsx
joyce-shi: i'm okay with it either way! was just curious.

although, in general, i don't think we should worry about # of files? i feel like we make sub-directories quite liberally & reorganize our folder structure frequently, so definitely feel free to do so as well! (we're definitely overdue for some cleaning 🧹). afaik, no drawbacks to that? let me know, if i'm wrong haha.

also, if the directory is not yet large, instead of preemptively trying to create less files, i feel like we can just organize when we get to that point. just my 2 cents!

jfdoming: I did it this way because I'm working on other components in // and it was getting kinda large lol... But yes, fair enough. I'll make a note of this for myself and come back to it in a separate PR once I have more components to organize.

PR #351, opened 2023-05-26
feat: add loading state to student login

joyce-shi:
## Notion ticket link

[Add loading state to student login](https://www.notion.so/uwblueprintexecs/Add-loading-state-to-Student-Login-b816424fe1184c4fbbe95e331756997a?pvs=4)

## Implementation description
* in cases where loading took a bit longer, it looked pretty weird w/o a loading state. added a quick impromptu one (i.e the spinner we use in our other loading state, but smaller). 
it also looks like there's been some linting bundled in here? not sure why it wasnt linted before though 🤔


frontend/src/components/pages/student/StudentLoginPage.tsx
```diff
@@ -62,6 +71,15 @@ const StudentLoginPage = (): React.ReactElement => {
   const image = STUDENT_SIGNUP_IMAGE;
   const form = (
     <>
+      {loading && (
+        <Spinner
+          color="blue.300"
+          emptyColor="gray.200"
+          size="md"
+          speed="0.65s"
+          thickness="4px"
```
jfdoming: Instead of adding all these props in both places (`LoadingState`), it might be nice to add these directly to the spinner `baseProps`/`default` variant in the theme? That way if we ever need to use a spinner in the future it will come pre-styled. Of course the `size` prop would still have to be set each time.

jfdoming (approved):
🙌

frontend/src/components/pages/student/StudentLoginPage.tsx
joyce-shi: good idea!

joyce-shi: oops just kidding. i think we actually can't set themes for certain props, i.e the ones that are not "stylistic": https://github.com/chakra-ui/chakra-ui/issues/1424

i'll make a custom component instead (:

jfdoming: Rip oh well, sounds like a good solution!

PR #356, opened 2023-05-28
fix: reload prompt

joyce-shi:
## Notion ticket link

[Fix reload issues](https://www.notion.so/uwblueprintexecs/Fix-reload-issues-286ed4eb8d9146269e58522ad3e59aee?pvs=4)

## Implementation description
* added reload prompt to student assessment experience
* fixed previous reload prompt (i.e it wasn't clearing when you navigated to other pages, so other pages would show this prompt)

* can i put the `useEffect` logic in a common hook like this? 👀


frontend/src/components/common/useReloadPrompt.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from "react";
+
+const useReloadPrompt = () => {
+  useEffect(() => {
+    window.onbeforeunload = () => true;
```
jfdoming: It's probably fine, but teeeeeeechnically this might not work if multiple components on the same page use this hook at once. Usually you'd prefer the `addEventListener`/`removeEventListener` API instead, but I'm fine leaving it as is since it's a bit of an edge case.

jfdoming (approved):

frontend/src/components/common/useReloadPrompt.ts
joyce-shi: you rite

PR #369, opened 2023-05-31
fix: Show correct login error 

Nimasheth:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* 

## Steps to test
1.

* 

## Checklist


frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -85,8 +84,7 @@ const registerTeacher = async (
       localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(user));
     }
   } catch (e: unknown) {
-    // eslint-disable-next-line no-alert
-    window.alert("Failed to sign up");
+    throw new Error("Failed to login");
```
carissa-tang:
```suggestion
    throw new Error("Failed to sign up");
```

frontend/src/components/auth/Login.tsx
```diff
@@ -26,31 +26,35 @@ const Login = (): React.ReactElement => {
...
-    if (user?.emailVerified === false) {
-      setUnverifiedUser(true);
-      return;
+
+    try {
+      const user: VerifiableUser | null = await authAPIClient.login(
+        email,
+        password,
+        login,
+      );
```
carissa-tang, line 53: we still need lines 49-51 here to check if the user is unverified

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
```diff
@@ -64,6 +67,9 @@ const TeacherSignupOne = ({
       default:
         break;
     }
+    if (field === "email") {
+      acceptEmailError(false);
```
carissa-tang: you can directly put this after line 62 in the switch case block

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
```diff
@@ -35,13 +35,16 @@ const TeacherSignupOne = ({
...
   const [emailError, setEmailError] = React.useState(false);
+  const [acceptEmail, acceptEmailError] = React.useState(false);
   const [gradesError, setGradesError] = React.useState(false);
   const history = useHistory();
 
   const { value: gradesValues, setValue: setGradesValue } = useCheckboxGroup({
     defaultValue: watch("grades") || [],
   });
 
+  const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
carissa-tang, line 46:
```suggestion
  const emailRegex = /\S+@\S+\.\S+/;
```
simplified

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
```diff
@@ -73,31 +79,51 @@ const TeacherSignupOne = ({
...
+    const emailValue = watch("email");
+    if (!emailValue || !!errors.email) {
       setEmailError(true);
-      return false;
+      isValid = false;
+    } else {
+      setEmailError(false);
+    }
+
+    if (!watch("email") || (emailValue && !emailRegex.test(emailValue))) {
```
carissa-tang:
```suggestion
    if (!emailRegex.test(emailValue)) {
```
we don't need to check for the empty case again since this was done above

carissa-tang (commented):
i think you're still missing the teacher signup error handling

basically, if the actual sign up request fails (i.e. you try to sign up with an email that already has an existing account), then we want to display that error on step 4. otherwise we will get this error 

image

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
jfdoming: I believe that the simplified regex would accept things like `@@@@.@`. Not sure how strict we want to be with validation though, realistically speaking I don't think people would intentionally type something like this (though I could imagine a typo like `example@@example.com`)

carissa-tang: you're right, we can leave as is then

i think you can actually have multiple @ under some specific circumstance (if they're in quotations) but these would be strange emails

Nimasheth:
> i think you're still missing the teacher signup error handling
> 
> basically, if the actual sign up request fails (i.e. you try to sign up with an email that already has an existing account), then we want to display that error on step 4. otherwise we will get this error
> 
> image

ohhh okok I will make that change!

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -85,8 +83,9 @@ const registerTeacher = async (
       localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(user));
     }
   } catch (e: unknown) {
-    // eslint-disable-next-line no-alert
-    window.alert("Failed to sign up");
+    if (e instanceof Error && e.message.includes("duplicate key")) {
+      throw new Error("Email already exists");
+    }
```
carissa-tang:
```suggestion
    if (e instanceof Error && e.message.includes("duplicate key")) {
      throw new Error("Email already exists");
    } else {
       throw new Error("Failed to sign up");
    }
```

frontend/src/components/auth/teacher-signup/steps/TeacherSignUpOne.tsx
```diff
@@ -35,13 +35,16 @@ const TeacherSignupOne = ({
   const [firstNameError, setFirstNameError] = React.useState(false);
   const [lastNameError, setLastNameError] = React.useState(false);
   const [emailError, setEmailError] = React.useState(false);
+  const [acceptEmail, acceptEmailError] = React.useState(false);
```
carissa-tang:
```suggestion
  const [acceptEmailError, setAcceptEmailError] = React.useState(false);
```
can you rename

carissa-tang (approved):

PR #375, opened 2023-05-31
feat: choose assessment page (distribute assessment - page 1)

joyce-shi:
## Notion ticket link

[Choose Assessment page (Distribute Assessment - Page 1)](https://www.notion.so/uwblueprintexecs/Choose-Assessment-page-Distribute-Assessment-Page-1-12f46a6107c34a9d8401adb3f6a4ab49?pvs=4)


frontend/src/themes/components/radio.ts
```diff
@@ -23,6 +23,15 @@ const Radio = {
...
     },
+    table: {
+      control: {
+        borderColor: "grey.300",
+        _checked: {
+          backgroundColor: "grey.300",
+          borderColor: "grey.300",
+        },
+      },
+    },
```
joyce-shi, line 34: I made this grey instead of blue like in the design, since if you make it blue, the selected radio becomes invisible on hover.

frontend/src/components/teacher/session-creation/steps/ChooseAssessment.tsx
```diff
@@ -1,23 +1,145 @@
...
+              assessments={assessments}
+              selectedTestId={testId}
+              setTestId={setTestId}
+            />
+          </VStack>
+        )}
+      </Box>
     </VStack>
   );
 };
```
joyce-shi, line 145: This is mainly a duplicate of the other assessments table, I'll come back to refactor when I have time (:

backend/graphql/resolvers/testResolvers.ts
```diff
@@ -113,6 +113,9 @@ const testResolvers = {
     tests: async (): Promise<TestResponseDTO[]> => {
       return testService.getAllTests();
     },
+    publishedTests: async (): Promise<TestResponseDTO[]> => {
```
jfdoming, line 116: I have a slight preference personally for adding an optional query argument `published: Boolean` to the existing mutation rather than creating a new one, but this should be okay too. The advantage of the optional argument is that we can reuse some of the same logic.

frontend/src/components/common/navigation/FormBreadcrumb.tsx
```diff
@@ -27,7 +27,7 @@ const FormBreadcrumb = ({
   validPage,
 }: FormBreadcrumbProps): React.ReactElement => {
   return (
-    <Breadcrumb color="grey.300" separator={<ChevronRightIcon />} spacing="8px">
+    <Breadcrumb color="grey.300" separator={<ChevronRightIcon />} spacing="2">
```
jfdoming: Nit (optional): when using chakra-sizes it can be nice to use numbers explicitly rather than strings to indicate the difference between the two types of sizes.

frontend/src/themes/components/table.ts
```diff
@@ -11,6 +11,7 @@ const Table = {
       },
       tr: {
         _hover: { backgroundColor: "blue.300", color: "grey.50" },
+        height: "52px",
```
jfdoming, line 14: This doesn't impact the other tables, right?

frontend/src/components/common/table/Table.tsx
```diff
@@ -52,8 +54,10 @@ export const Table = <T extends Nodes = Nodes>({
           <Tbody>
             {paginatedData.map((row, rowIndex) => (
               <Tr
-                key={rowIndex}
+                key={row.id || rowIndex}
```
jfdoming, line 57: Just a thought: defaulting to the index for a common component like a table could be dangerous if we're navigating between pages. Could also be fine though depending on the circumstances

jfdoming (approved):
Disliking how large our tables are getting, but I don't think there's an easy way around it. And not worth solving given our time constraints. LGTM!

backend/graphql/resolvers/testResolvers.ts
joyce-shi: good point! updating now

joyce-shi:
i'm merging this now, because it's been open too long and i cant look at it anymore. i'll come back and apply fixes when i have more mental capacity for it lol

PR #377, opened 2023-06-01
fix: optimize test session resolvers

jfdoming:
## Notion ticket link

[Optimize test session resolvers](https://www.notion.so/uwblueprintexecs/Optimize-test-session-resolvers-5841edee079345f89a139525da870fbb)

## Implementation description
* Converted the `testSession` and `class` service-resolver pairs to use the resolver-chain pattern rather than the `mapToDTOs` pattern

## Steps to test
1.

* 

## Checklist


jfdoming:
There shouldn't be any more issues with this PR since we're using built-in mongoose JSON serialization, but feel free to check out the PR locally and test a few features if you like.

backend/graphql/resolvers/classResolvers.ts
```diff
@@ -26,7 +26,6 @@ const classService: IClassService = new ClassService(
   userService,
   testSessionService,
 );
-testSessionService.bindClassService(classService);
```
jfdoming, line 29: It makes me so happy to delete this method lol

backend/models/index.ts
```diff
@@ -9,6 +9,7 @@ export const mongo = {
       return;
     }
     try {
+      mongoose.set("toObject", { virtuals: true });
```
jfdoming, line 15: This basically just tells mongoose to serialize the `id` field

backend/graphql/resolvers/classResolvers.ts
joyce-shi: LOL

backend/graphql/resolvers/classResolvers.ts
```diff
@@ -67,6 +66,14 @@ const classResolvers = {
...
     },
   },
+  ClassResponseDTO: {
+    teacher: async (parent: ClassResponseDTO) => {
+      return userService.getUserById(parent.teacher);
+    },
+    testSessions: async (parent: ClassResponseDTO) => {
+      return testSessionService.getTestSessionsByClassId(parent.id);
+    },
+  },
```
joyce-shi, line 76: wait, i'm not entirely sure what's going on here. does this overload the existing imported `ClassResponseDTO`? or how does it work

backend/graphql/resolvers/testSessionResolvers.ts
```diff
@@ -30,75 +29,62 @@ const classService: IClassService = new ClassService(
...
+    test: ({ test }: TestSessionResponseDTO) => testService.getTestById(test),
+    results: async ({
+      class: parentClass,
+      results,
+    }: TestSessionResponseDTO) => {
+      if (!results) return [];
+
+      // Returning all students whenever results are requested is not ideal, but
+      // it's better than making the same request for each result when the
+      // students are needed.
```
joyce-shi: i think it should be fine no? cuz there aren't use cases where we request specific students for now

backend/services/interfaces/testSessionService.ts
```diff
@@ -34,18 +29,18 @@ export interface TestSessionResponseDTO {
   /** the unique identifier for the test session */
   id: string;
   /** the corresponding test from the Test collection */
-  test: TestResponseDTO;
+  test: string;
   /** the teacher administering the test from the User collection */
```
joyce-shi: nit: (& same for entries below)
```suggestion
  /** the id of the teacher administering the test from the User collection */
```

joyce-shi (approved):
looks fantastic. just left a question, mainly for my own learning (:

backend/graphql/resolvers/testSessionResolvers.ts
jfdoming: Tru

backend/services/interfaces/testSessionService.ts
jfdoming: Copilot not yet smart enough to generate fully correct comments, I see 🤓

backend/graphql/resolvers/classResolvers.ts
jfdoming: Yeah, exactly. So the way I set it up is that the main resolver returns a DTO with just the ID for this field. And then there are two cases: if the frontend requested the `testSessions` field, then Apollo will additionally run this resolver, and the return value replaces the original ID; otherwise, the ID is just discarded without running the extra resolver. So the net effect is that we only hit the database an extra time if the field is actually requested, which seems to shave a TON of time off the response time.

backend/graphql/resolvers/testSessionResolvers.ts
```diff
@@ -30,75 +29,62 @@ const classService: IClassService = new ClassService(
...
-            student: studentDTO,
-          };
-        }),
-      };
-    },
-    testSessions: async (): Promise<TestSessionResponseDTO[]> => {
-      return testSessionService.getAllTestSessions();
-    },
-    testSessionByAccessCode: async (
+    testSession: (_parent: undefined, { id }: { id: string }) =>
```
carissa-tang, line 35: wondering why we removed the return values? is it just cause its verbose or there's actual code difference?

backend/models/index.ts
carissa-tang, line 15: can you add a brief comment here describing what it does

backend/services/interfaces/testSessionService.ts
```diff
@@ -34,18 +29,18 @@ export interface TestSessionResponseDTO {
...
+  school: string;
   /** the class taking the test session */
-  class: ClassResponseDTO;
+  class: string;
   /**
    * the result of the test session
    * there should be one entry here per student
    */
-  results: ResultResponseDTO[];
+  results?: ResultResponseDTO[];
```
carissa-tang, line 43: shouldn't the response always have a results field? (but could be `[]`)

backend/utilities/generalUtils.ts
```diff
@@ -25,3 +27,9 @@ export const isCompletedTestResult = (
 ): boolean => {
   return results.flat().every((answer) => answer.length);
 };
+
+export const mapDocumentToDTO = <T extends Document>(document: T) =>
```
carissa-tang, line 31: this is 👌👌 so clean

carissa-tang (commented):
this is greattt ty for doing this

carissa-tang (commented):
this is greattt ty for doing this

backend/graphql/resolvers/testSessionResolvers.ts
jfdoming: Mainly verbosity, yeah. Since we're returning the value of the service directly in basically all cases, it's just a duplicate type essentially

backend/services/interfaces/testSessionService.ts
jfdoming: My understanding is that it could be missing from Mongo objects. Previously, we manually filled in this field before returning it from the service, but I think it's more correct to have GraphQL handle this sort of defaulting (hence the fact that the TypeScript type makes the field optional; the GraphQL type still has it as a required field). But I'm open to alternatives depending on your thoughts here.

PR #378, opened 2023-06-02
fix: crashing heroku deployment

jfdoming:
## Notion ticket link
n/a

## Implementation description
- [fix: Copy .cjs files to build/ directory after yarn build](https://github.com/uwblueprint/jump-math/pull/378/commits/2291ae1ba067a77a05ee1269402a3d9aea83af86)
- [fix: Remove the need for .cjs extensions for library files](https://github.com/uwblueprint/jump-math/pull/378/commits/be1d1a01cc303d6a30717f39054f83a75fedf9bc)
- [fix: Add eslint rule to catch manual imports of .cjs files](https://github.com/uwblueprint/jump-math/pull/378/commits/3ad3708f9156192c47226b75edcb67da8f2109da)
- [fix: Ignore build directory when using nodemon](https://github.com/uwblueprint/jump-math/pull/378/commits/9b54a78fda1eedca8a4bdc95a1d9032d688c2859)
- [fix: Lint ALL project files (not just .ts/.js files)](https://github.com/uwblueprint/jump-math/pull/378/commits/4bf4e5a9b80d09a879d500edfe15d6a26af6c1e6)

## Steps to test
0. Consider `rm -rf build/` for a fresh start.
1. Run `yarn build` locally to generate your `build/` directory.
2. Run `node build/server.js` and verify that server does not crash due to an import error (it might crash due to missing environment variables).

* 

## Checklist


jfdoming:
Explanation: we're using the `graphql-upload` library for file uploads. However, the maintainer of that library made a decision (that I hate) to use CommonJS modules in their project in newer versions (and we updated to this newer version in order to fix a security vulnerability that actually had an impact unfortunately). CommonJS modules are like a virus (metaphorically) in that if you use a single CJS dependency, then normally you'd have to make your entire project CJS. Unfortunately, CJS is straight-up incompatible with a lot of other libraries we use, so that's not an option.

I solved this problem a while back by manually generating a few `.cjs` files and importing them, which works fine in development. However, the TypeScript compiler doesn't know about files with that extension by default, so at build time (which only occurs on Heroku), it was failing to copy those `.cjs` files to the `build/` directory, resulting in runtime errors. The solution is 1) to instruct TypeScript to copy all JavaScript files in the project into the `build/` directory via the `allowJs` option, and 2) to rename our `.cjs` files to `.js` files so that TypeScript will become aware of them (it doesn't know about the `.cjs` extension).

It's ugly, but it works, so...

joyce-shi (approved):
thank you for fixing! 😎 🔨 ( <- this is a picture of you)

jfdoming:
lmao np. Unfortunately though this "fix" doesn't actually work in all cases (there are like 5 different things I need to check in order to make sure it works) 💀 working on a better fix, but let's not merge yet plz

also it makes me sad that our tests didn't catch any of this 😭

jfdoming:
@joyce-shi the new fix is to manually copy all `.cjs` files from the `lib/` directory into `build/lib`. Not the best solution, but the best solution would be to migrate to an actual build tool like `babel` rather than using TypeScript directly, and that would be way more of a hassle.

backend/package.json
```diff
@@ -26,6 +27,7 @@
     "dotenv": "^16.0.3",
     "eslint-config-airbnb-base": "15.0.0",
     "express": "^4.18.2",
+    "fast-glob": "^3.2.12",
```
jfdoming, line 30: This dependency was already being installed by other packages, so adding it here won't affect our usage at all!

joyce-shi (approved):

PR #353, opened 2023-05-27
feat: fetch data for classroom page

d282liu:
## Notion ticket link
[Fetch data for Classrooms page](https://www.notion.so/uwblueprintexecs/Fetch-data-for-Classrooms-page-0c818d389ee24cd0ab51f1d14c0ad94b?pvs=4)

## Implementation description
* Fetched data for classroom page using UseQuery

## Notes
* Some of the fields right now for the classroom cards are hard coded - I will be addressing that in a future PR by creating a new type and making changes on the backend. I will create the ticket and assign to myself at some point 

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have run the appropriate linter(s)
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #393, opened 2023-06-05
feat: fetch data for classroom page without hardcoded fields

d282liu:
Notion ticket link
[Fetch data for Classrooms page](https://www.notion.so/uwblueprintexecs/Fetch-data-for-Classrooms-page-0c818d389ee24cd0ab51f1d14c0ad94b?pvs=4)

Implementation description
Fetched data for classroom page using UseQuery, replaced hard coded values in previous PR (https://github.com/uwblueprint/jump-math/pull/353) with actual data by creating a new type `ClassCard`

Notes
- Not too sure right now how to set the `activeAssessments` field

## Checklist


backend/graphql/resolvers/classResolvers.ts
```diff
@@ -39,8 +40,22 @@ const classResolvers = {
...
+        classCards.push({
+          id: classObj.id,
+          activeAssessments: classObj.testSessions.length,
+          assessmentCount: classObj.testSessions.length,
+          gradeLevel: classObj.gradeLevel,
+          className: classObj.className,
+          studentCount: classObj.students.length,
+        });
+      });
+      return classCards;
```
jfdoming:
```suggestion
      const classesByTeacher = await classService.getClassesByTeacherId(teacherId);
      return classesByTeacher.map((classObj) => ({
        ...classObj,
        activeAssessments: classObj.testSessions.length,
        assessmentCount: classObj.testSessions.length,
        studentCount: classObj.students.length,
      }));
```

jfdoming (approved):
Just one comment! Also, if you decide to merge all the props from `ClassCard` into `ClassResponseDTO` like we discussed on Discord, then it would be better to do this mapping within the service itself. (Or, if you want to be extra extra fancy, you can use [virtual fields](https://mongoosejs.com/docs/tutorials/virtuals.html) on the `mongoose` model and everything should magically work!)

backend/graphql/resolvers/classResolvers.ts
d282liu: Sounds good! I think I will maybe merge all the props for `ClassCard` into `ClassResponseDTO` in a later ticket if we find that the query `classesByTeacher` in the resolvers is needed elsewhere and requires a `ClassResponseDTO`, or if the fields for `ClassCard` are needed in other queries. Thank you for the suggestions

jfdoming (approved):

PR #380, opened 2023-06-03
(1) feat: fetch data for visualization

jfdoming:
## Notion ticket link

[Fetch data for data-viz page](https://www.notion.so/uwblueprintexecs/Fetch-data-for-data-viz-page-858dcbd4119b41bc91bf443fc3f50ba1)

## Implementation description
- Load test session title from backend
- Load student list & scores from backend

## Steps to test
1. View the data-viz page for an assessment with at least one result.

## What should reviewers focus on?
* 

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #397, opened 2023-06-07
(2) feat: compute percentile statistic

jfdoming:
## Notion ticket link

[Add percentile scoring](https://www.notion.so/uwblueprintexecs/Add-percentile-scoring-311d6043a67d4e79ae489a295fd6f716)

## Implementation description
* Follow-up to #380 

## Steps to test
1. View an assessment drilldown and verify you see the correct percentile.

* Clean? Correct?

## Checklist

### DIFF
--- a/backend/graphql/resolvers/testSessionResolvers.ts
+++ b/backend/graphql/resolvers/testSessionResolvers.ts
@@ -13,6 +13,7 @@ import type IUserService from "../../services/interfaces/userService";
 import type { ISchoolService } from "../../services/interfaces/schoolService";
 import ClassService from "../../services/implementations/classService";
 import type { IClassService } from "../../services/interfaces/classService";
+import { computePercentiles } from "../../utilities/generalUtils";
 
 const userService: IUserService = new UserService();
 const schoolService: ISchoolService = new SchoolService(userService);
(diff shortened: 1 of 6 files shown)


backend/graphql/resolvers/testSessionResolvers.ts
```diff
@@ -68,7 +69,7 @@ const testSessionResolvers = {
       if (!results) return [];
 
       const { students } = await classService.getClassById(parentClass);
-      const resultsByStudentId = results.reduce<
+      const resultsByStudentId = computePercentiles(results, "score").reduce<
```
joyce-shi, line 72: do we re-compute percentile every time we query for student results?

joyce-shi (commented):
just had one q

backend/graphql/resolvers/testSessionResolvers.ts
jfdoming: Yes, I had remembered that we were deferring optimization until later. It's just a sort so I'm not too too worried about performance, but ultimately we should probably move it to the completion mutation or something. Do you think we should do that now?

joyce-shi: all good. approving!

joyce-shi (approved):

PR #460, opened 2023-07-16
fix: classroom card spacing

joyce-shi:
## Notion ticket link

[Fix spacing on classroom cards](https://www.notion.so/uwblueprintexecs/Fix-Spacing-on-Classroom-Cards-a646025e969c485db2fcba97b8c3f93e?pvs=4)

## Implementation description
* Using Flexbox to adapt the # of columns based on screen size
[i.e](https://drive.google.com/file/d/13hSZ6NJ1347nrTHsP8-F3Rkuv4UyFAtG/view?usp=sharing)

## Steps to test
1. UI renders as expected

### DIFF
--- a/frontend/src/components/pages/teacher/ClassroomsPage.tsx
+++ b/frontend/src/components/pages/teacher/ClassroomsPage.tsx
@@ -5,8 +5,7 @@ import { useQuery } from "@apollo/client";
 import {
   Box,
   Center,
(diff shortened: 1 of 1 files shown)


jfdoming (approved):
Surprised this was needed since IIRC this is exactly the case CSS grid is supposed to handle, but if it works it works 👍

PR #461, opened 2023-07-16
feat: choose classroom page (distribute assessment - page 2)

joyce-shi:
## Notion ticket link

[Choose Classroom page](https://www.notion.so/uwblueprintexecs/Choose-Classroom-page-Distribute-Assessment-Page-2-4de893e575f44d98af081c231ca29451?pvs=4)

## Implementation description
* Added the second page in the Distribute Assessment flow
There are some design liberties I took here, such as empty states and hover behaviour, but I think we can sync up with design on this when we do a full review.

One key difference - I left the search bar off of this page. We don't implement a search bar for the Classrooms page, so I don't think it makes sense to implement one here. I think if we do want to, we should align to add this search bar to both views, so leaving it off this one for now.


frontend/src/components/common/info/messages/EmptyDistributeClassroomsMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+        onClick={onClick}
+        paragraphs={["You currently have no classrooms."]}
+        subtitle="Please navigate to the Classrooms page to add a classroom."
+        textColor="blue.300"
+      />
+    </Center>
+  );
+};
+
+export default EmptyDistributeClassroomsMessage;
```
joyce-shi, line 31: Not specced by design but took some liberties here.

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
```diff
@@ -1,16 +1,146 @@
...
+                  pagesCount={totalPages}
+                />
+              )}
+            </VStack>
+          </Box>
+        ) : (
+          <EmptyDistributeClassroomsMessage onClick={navigateToClassrooms} />
+        )}
+      </Box>
+    </VStack>
```
joyce-shi: some of this is duplicate from `ClassroomPages`, but I didn't think it was worth separating into a parent component (since we only use this pattern twice for now), but let me know if you think we should. Or, if there are some ways we can separate things into common components.

frontend/src/components/teacher/student-management/classrooms/ClassroomCard.tsx
```diff
@@ -72,18 +83,22 @@ const ClassroomCard = ({
...
+                to={{
+                  pathname: Routes.DISPLAY_CLASSROOM_PAGE(id),
+                  state: { className: name, startDate, grade },
+                }}
+              >
+                {classroomTitle}
+              </LinkOverlay>
+              <ClassroomPopover />
+            </>
+          )}
```
joyce-shi: Same here, I didn't think the differences warranted a separate component, so just tried to make this one accommodate the minor differences from this view.

frontend/src/components/teacher/student-management/classrooms/ClassroomCard.tsx
```diff
@@ -28,6 +28,8 @@ interface ClassroomCardProps {
   assessmentCount: number;
   grade: Grade;
   activeAssessments: number;
+  disabledOptions?: boolean;
```
carissa-tang: maybe rename to something more descriptive like `clickDisabled`? or were there other event actions you had in mind under "options"?

frontend/src/components/common/info/messages/EmptyDistributeClassroomsMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type EmptyDistributeClassroomsStateProps = {
+  onClick?: () => void;
+};
+
+const EmptyDistributeClassroomsMessage = ({
+  onClick,
+}: EmptyDistributeClassroomsStateProps): React.ReactElement => {
+  return (
+    <Center
```
carissa-tang: we have a lot of repeats of this in the codebase (like 5) - could we make a theme variant for `<Center>`?

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
```diff
@@ -1,16 +1,146 @@
...
+      ({ testSessions, students, ...classCard }) => {
+        let activeAssessments = 0;
+        testSessions.forEach((session) => {
+          if (
+            getSessionStatus(session.startDate, session.endDate, now) ===
+            "active"
+          ) {
+            activeAssessments += 1;
+          }
+        });
```
carissa-tang, line 52: hmm this is my bad, but we should prob handle this in the resolver to keep it cleaner. or, can we at least refactor into a helper function? (for the duplicate with ClassroomsPage)

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
```diff
@@ -1,16 +1,146 @@
...
+  }, [data]);
+
+  const { paginatedData, totalPages, currentPage, setCurrentPage } =
+    usePaginatedData(classCards);
+
   return (
-    <Text color="blue.300" textAlign="left" textStyle="header4">
-      Choose a Classroom
-    </Text>
+    <VStack align="left" spacing="2">
```
carissa-tang: do you think it's worth creating a wrapper for distribute flow? with the header, text and loading / error states

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
carissa-tang: hmm yea there's a lot of minor differences that'll make the code ugly so i think fine as is

carissa-tang (commented):
lookin good! left a few comments

frontend/src/components/common/info/messages/EmptyDistributeClassroomsMessage.tsx
joyce-shi: good ideal! will do

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
joyce-shi: okay to do this in a follow-up PR? https://www.notion.so/uwblueprintexecs/Add-status-past-active-future-field-to-resolver-cbe782a6807d425c8262dfb2ed630f00?pvs=4

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
joyce-shi: donee

frontend/src/components/common/info/messages/EmptyDistributeClassroomsMessage.tsx
joyce-shi: implemented this according to: https://github.com/chakra-ui/chakra-ui/issues/2456

let me know if there's a more straightforward way of doing this 🤔

carissa-tang: ah shame.. should be ok then. ig one other way is to make a custom `<Center>` component

frontend/src/components/pages/teacher/DisplayAssessmentsPage.tsx
```diff
@@ -78,23 +79,26 @@ const DisplayAssessmentsPage = (): React.ReactElement => {
...
         </Center>
       )}
       {error && (
-        <Box height="100%" mt={10}>
+        <Box>
           <ErrorState />
         </Box>
       )}
-      {!!formattedData?.length && !loading && !error && (
+      {!!formattedData?.length && !loading && error && (
```
carissa-tang:
```suggestion
      {!!formattedData?.length && !loading && !error && (
```
is it supposed to be this?

frontend/src/components/teacher/session-creation/steps/ChooseClass.tsx
carissa-tang: ya sg

carissa-tang (approved):
lgtm! one small comment, but approving to unblock you

frontend/src/components/teacher/session-creation/DistributeAsessmentWrapper.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
```
carissa-tang, line 1: wait also can you fix the typo in the filename 🫣

carissa-tang (changes requested):
actually wait, it looks like the search count is wrong. looks like we're taking the [length](https://github.com/uwblueprint/jump-math/pull/461/files#diff-bca331c11e3088634acf2045418536d0dbecd4a1cb038f6e28233bd5a630f8eaR129) of the search keyword, not the results


image

joyce-shi:
> actually wait, it looks like the search count is wrong. looks like we're taking the [length](https://github.com/uwblueprint/jump-math/pull/461/files#diff-bca331c11e3088634acf2045418536d0dbecd4a1cb038f6e28233bd5a630f8eaR129) of the search keyword, not the results
> 
> image

that code isn't implemented in the scope of this PR, so i think the better practice here is to create a bug ticket for this on the task board. could you do that?

carissa-tang (approved):

PR #478, opened 2023-07-23
feat: distribute assessment submission

joyce-shi:
## Notion ticket link

[Distribute Assessment Submission](https://www.notion.so/uwblueprintexecs/Distribute-Assessment-Submission-7b5a3d79e0e742d2bc06b9f5f9910258?pvs=4)

## Implementation description
* Implemented frontend mutation to submit assessment

Note: this differs from design. Once the test is created, it just redirects the user to the table of test sessions. This is the same design pattern that we use in the other parts of the application and I don't want to start creating new screens now 👍. We can confirm this with design later.

## Steps to test
1. Works as expected


frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -123,14 +128,22 @@ const DistributeAssessmentPage = (): React.ReactElement => {
...
-          >
-            Distribute
-          </Button>
+          <>
+            {startDate && endDate && teacherId && (
+              <DistributeAssessmentButton
+                testSessionRequest={{
+                  test: testId,
+                  teacher: teacherId,
+                  school: "639151a4d46e8c002a49f8d6",
```
joyce-shi, line 136: Right now, each `TestSession` has a `school` field. But I have difficulty here populating this field, because we don't actually link `teacher` to `school`. I would have to add a new query `GET_SCHOOL_BY_TEACHER` (and check each school's array of teachers if a given `teacher` belongs to it). 

But I think this begs the question of whether we should just be storing the `school` in the `teacher` object to begin with? i.e why do we store the `school` as a reference in the `test session` object but not in the `teacher` object?

joyce-shi:
also, I noticed that we have a `GET_SCHOOL` mutation but we never use it. any idea why?

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
joyce-shi: Based on Slack convo, I'll add `school` field to `teacher`, but I'll do so in a separate PR so the scope of this one doesn't grow too large! for now, the value will be hardcoded.

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -17,6 +19,9 @@ const BREADCRUMB_CONFIG: BreadcrumbType[] = [
 ];
 
 const DistributeAssessmentPage = (): React.ReactElement => {
+  const { authenticatedUser } = useContext(AuthContext);
+  const { id: teacherId } = authenticatedUser ?? {};
```
jfdoming, line 22: This is such a common pattern I feel like we should have created a hook for this a while back 😆 but all good now

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const [createSession, { loading, error }] = useMutation<{
+    createSession: string;
+  }>(CREATE_TEST_SESSION, {
+    refetchQueries: [{ query: GET_TEST_SESSIONS_BY_TEACHER_ID }],
+  });
+
+  const handleCreateSession = async (testSession: TestSessionRequest) => {
+    await createSession({ variables: { testSession } });
+    if (error) {
```
jfdoming: Does this work? I feel like with the way closures work it might not as written. If you haven't yet, maybe consider throwing an error on the backend and testing that way? Usually the pattern would be to use try/catch around the `await`

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface DistributeAssessmentButtonProps {
+  testSessionRequest: TestSessionRequest;
+}
+
+const DistributeAssessmentButton = ({
+  testSessionRequest,
+}: DistributeAssessmentButtonProps): React.ReactElement => {
+  const history = useHistory();
+  const { showToast } = Toast();
```
jfdoming: this is the nittiest of nits and not related to your pr, but any chance we could rename `Toast` to `useToast`? It's a custom hook but we're not using a custom-hook name name for it and that's been driving me crazy 😆 (for a more concrete reason why it might be nice, this gets us linter rules which check for rules-of-hooks, which can help catch bugs)

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
```diff
@@ -123,14 +128,22 @@ const DistributeAssessmentPage = (): React.ReactElement => {
...
-          </Button>
+          <>
+            {startDate && endDate && teacherId && (
+              <DistributeAssessmentButton
+                testSessionRequest={{
+                  test: testId,
+                  teacher: teacherId,
+                  school: "639151a4d46e8c002a49f8d6",
+                  class: classId,
+                  accessCode: generateAccessCode(),
```
jfdoming: This will regenerate the access code every time this component renders, which react can choose to do at any point. Is that fine?

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
joyce-shi: lol sure

frontend/src/components/pages/teacher/DistributeAssessmentPage.tsx
joyce-shi: u right, dats probs bad

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    createSession: string;
+  }>(CREATE_TEST_SESSION, {
+    refetchQueries: [{ query: GET_TEST_SESSIONS_BY_TEACHER_ID }],
+  });
+
+  const handleCreateSession = async () => {
+    try {
+      await createSession({
+        variables: {
+          testSession: { ...testSession, accessCode: generateAccessCode() },
```
jfdoming: my instinct would be to put this on the backend actually but I can't think of a concrete reason why. I suppose a bad actor could probably break our infra by spamming the same code over and over but I think this is ok

jfdoming (approved):

PR #489, opened 2023-07-24
feat: add fraction component

carissa-tang:
## Notion ticket link

[Add fraction as component as question element](https://www.notion.so/uwblueprintexecs/Add-fraction-as-component-as-question-element-96bea76669524cbd96cc606b679e5000?pvs=4)

## Implementation description
Updated the following experiences to include fractions:
* **Admin**: question editor (creating and editing a fraction), fraction tag in question card
* **Teacher**: displaying student results
* **Student**: assessment summary page, student test taking

Notes:
- Aligned fraction in center so the expanding width can be aligned
- Will implement mixed fractions in follow up PR
- Will refactor common code among other question elements in follow up PR
- Will replace question images with question icons in follow up PR

## Steps to test

Student experience

https://github.com/uwblueprint/jump-math/assets/69697180/f449daf5-4c32-4de1-8ffc-50fec1d2c6c2

image

Teacher experience
image

Admin experience
image

https://github.com/uwblueprint/jump-math/assets/69697180/5b42f9fc-67b6-4330-92b5-c73270bd0c4b

* 

## Checklist


frontend/src/assets/images.ts
```diff
@@ -34,3 +34,8 @@ export const SHORT_ANSWER: ImageType = {
   src: "https://storage.googleapis.com/jump-math-98edf.appspot.com/short-answer.png",
   alt: "short-answer",
 };
+
+export const FRACTION: ImageType = {
+  src: "https://storage.googleapis.com/jump-math-98edf.appspot.com/fraction.png",
```
jfdoming, line 39: Side note but do we have a ticket to deal explicitly with setting these up in prod? I feel like this would be an easy thing to miss once we spin down staging.

frontend/src/components/admin/question-creation/question-elements/modals/fraction/AddFractionModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { v4 as uuidv4 } from "uuid";
+
+import QuestionEditorContext from "../../../../../../contexts/QuestionEditorContext";
+import type { FractionMetadata } from "../../../../../../types/QuestionMetadataTypes";
+import { QuestionElementType } from "../../../../../../types/QuestionTypes";
+
+import FractionModal from "./FractionModal";
+
+const AddFractionModal = (): React.ReactElement => {
+  const { onClose } = useDisclosure();
```
jfdoming: nit: [since `useDisclosure` works kind of like a fancy `useState` anyway](https://github.com/chakra-ui/chakra-ui/blob/main/packages/hooks/use-disclosure/src/index.ts#L20), I don't think calling their `onClose` is really necessary here. You could probably just call `setShowAddFractionModal(false)` and be done with it

frontend/src/components/teacher/results/StudentAnswersSection/question-elements/CorrectedFraction.tsx
```diff
@@ -0,0 +1,40 @@
...
+import CorrectedTextField from "./CorrectedTextField";
+
+type CorrectedFractionProps = {
+  studentAnswer?: (number | undefined)[];
+  correctAnswer: FractionMetadata;
+};
+
+const CorrectedFraction = ({
+  studentAnswer,
+  correctAnswer,
```
jfdoming, line 16: nit: some of the code below might be a little more legible if we destructured the student answer. Something like:
```
  const [studentAnswerNumerator, studentAnswerDenominator] = studentAnswer;
```

Just a personal preference, it's alright to keep as is though

frontend/src/constants/StudentAssessmentConstants.ts
```diff
@@ -1,22 +1,32 @@
...
   SHORT_ANSWER: {
     src: SHORT_ANSWER.src,
     alt: "short-answer",
-    tooltip: "Users will have to type in their answers",
+    tooltip: editorTooltips.SHORT_ANSWER,
+  },
+  FRACTION: {
+    src: FRACTION.src,
+    alt: "fraction",
+    tooltip: editorTooltips.FRACTION,
```
jfdoming: I'm thinking we might eventually want to change the language in the student-facing tooltips to something more intelligible to a Grade 1 student, like "Answer these questions using fractions (numerator and denominator)." Since that would probably differ from what we'd show a teacher, I'm thinking we should either extract a separate student-specific constant list or just revert to the previous form. Thoughts?

frontend/src/constants/QuestionConstants.ts
```diff
@@ -7,6 +7,7 @@ const editorTooltips = {
     "Users will have to select the correct responses from a list of options",
   MULTIPLE_CHOICE:
     "Users will have to select the best response from a list of options",
+  FRACTION: "Users will be able to input fractions in their answer",
```
jfdoming: I'm also thinking it would be good to update the language here slightly (change "Users" to "Students") since teachers/admins won't think of students as "users," but I'm fine to do that in a separate ticket to generally improve copy.

frontend/src/components/admin/question-creation/question-elements/modals/fraction/EditFractionModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+const EditFractionModal = ({
+  id,
+  data,
+  isOpen,
+  setOpen,
+}: EditFractionModalProps): React.ReactElement => {
+  const { onClose } = useDisclosure();
```
jfdoming: Similar comment about `useDisclosure` from before. In this case though I think it might be better to use a `useDisclosure` in the parent component (admin `FractionElement`) and pass down the `isOpen` and `onClose` props.

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+    width: number;
+  }): React.ReactElement => (
+    <Input
+      border="2px solid #636363"
+      borderRadius={2}
+      ml={1}
+      onChange={onChange} // CHECK
+      readOnly={readOnly}
+      textAlign="center" // CHECK
+      type="number" // CHECK
```
jfdoming: Just wondering, what is the purpose of these comments?

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  denominator,
+  numerator,
+  onNumeratorChange,
+  onDenominatorChange,
+  readOnly = false,
+}: FractionWrapperProps): React.ReactElement => {
+  const numeratorWidth = getFractionInputWidth(numerator);
+  const denominatorWidth = getFractionInputWidth(denominator);
+
+  const FractionInput = ({
```
jfdoming: nit: I'd prefer to define components at the top level, since the nesting can make it a bit hard to reason about things.

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  readOnly?: boolean;
+}
+
+const getFractionInputWidth = (value: string): number => {
+  return value.length > 1 ? 48 + 8 * value.length : 48;
+};
+
+const getDividerWidth = (width1: number, width2: number) => {
+  return Math.max(width1, width2) + 12;
+};
```
jfdoming: Would it be possible to do each of these as a CSS `calc` instead of a Javascript calculation, at least partially? With Chakra UI I've found it's safer to base things off the `rem` unit since that's what they use for everything. That way if we ever had to scale up the UI slightly things would still be the right size. If this is a ton of work though to get right I'm alright leaving it as is

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <FractionInput
+        onChange={onDenominatorChange}
+        value={denominator}
+        width={denominatorWidth}
+      />
+    </VStack>
+  );
+};
+
+export default FractionWrapper;
```
jfdoming: Could we rename this to `FractionInput` and have the other component called `FractionFieldInput` or something? `FractionWrapper` is a little unclear when searching the codebase IMO

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+    value,
+    onChange,
+    width,
+  }: {
+    value: string;
+    onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    width: number;
+  }): React.ReactElement => (
+    <Input
+      border="2px solid #636363"
```
jfdoming: nit: generally I'd prefer using `borderColor` and `borderWidth` so that we can use the colours from the scheme. That way if we ever had to update it we wouldn't have to manually change many places

frontend/src/components/common/FractionWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+      type="number" // CHECK
+      value={value}
+      width={`${width}px`}
+    />
+  );
+
+  return (
+    <VStack
+      alignItems="center"
+      border={readOnly ? "none" : "1px solid #636363"}
```
jfdoming: ditto on using `borderColor` and `borderWidth`

jfdoming (commented):
Left a few minor comments but looks great 💯

frontend/src/assets/images.ts
carissa-tang: good idea, made a ticket https://www.notion.so/uwblueprintexecs/Set-up-GCP-images-in-PROD-394f97d8c1b540128949d72402950c91?pvs=4

frontend/src/constants/StudentAssessmentConstants.ts
carissa-tang: true, will make copy changes

frontend/src/components/common/FractionWrapper.tsx
carissa-tang: LOL forgot to remove them...

frontend/src/components/common/FractionWrapper.tsx
carissa-tang: found a different way to handle it that's slightly better

frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
```diff
@@ -53,20 +53,28 @@ const AssessmentEditorHeader = ({
...
+  const {
+    onOpen: onDeleteModalOpen,
+    isOpen: isDeleteModalOpen,
+    onClose: onDeleteModalClose,
+  } = useDisclosure();
+  const {
+    onOpen: onArchiveModalOpen,
+    isOpen: isArchiveModalOpen,
+    onClose: onArchiveModalClose,
+  } = useDisclosure();
```
joyce-shi, line 70: nice

joyce-shi (approved):
looks good to me!

PR #495, opened 2023-07-27
fix: add contextually-appropriate titles to all pages

jfdoming:
## Notion ticket link

[Change page titles based on current page](https://www.notion.so/uwblueprintexecs/Change-page-titles-based-on-current-page-0e666fb3f9bf4c218062f1e2900fc00e)

## Implementation description
Titles are set using two possible methods. First, static titles can be provided to any `PrivateRoute` (and optionally marked as overridable by child components using `isTitleOverridable`). Second, dynamic titles can be set via a `usePageTitle` hook. Contextually-appropriate titles have been set on all pages that existed at the time of writing.

## Steps to test
1. View all pages in the application and check the tab title.

* Are there any titles you think could be more accurate?

## Checklist


frontend/src/components/pages/admin/AdminRouting.tsx
```diff
@@ -36,12 +38,14 @@ const AdminRouting = (): React.ReactElement => {
                 exact
                 path={Routes.USER_DATABASE_PAGE}
                 roles={["Admin"]}
+                title="Users"
```
carissa-tang: i feel like this should be `Database` to keep consistent with the page name - thought database is kind of confusing in the first place lol

frontend/src/components/pages/student/StudentRouting.tsx
```diff
@@ -61,12 +61,14 @@ const StudentRouting = (): React.ReactElement => {
...
             path={Routes.ASSESSMENT_SUMMARY_PAGE}
             roles={["Student"]}
+            title="Assessment Summary"
           />
           <PrivateRoute
             component={WriteAssessmentPage}
             exact
             path={Routes.WRITE_ASSESSMENT_PAGE}
             roles={["Student"]}
+            title="Write Assessment"
```
carissa-tang: why do we need both this line and the dynamic config in `WriteAssessmentPage`? (same thing with assessment summary)

frontend/src/components/pages/student/AssessmentSummaryPage.tsx
```diff
@@ -33,6 +34,8 @@ const AssessmentSummaryPage = (): React.ReactElement => {
     history.push(WRITE_ASSESSMENT_PAGE);
   };
 
+  usePageTitle(test?.name ?? "Assessment Summary");
```
carissa-tang, line 37: this is defaulting to `"Assessment Summary"` on initial load and only if i press start + go back then it shows the test name. it's a small detail but just wondering why that is.

would it work if we pass in `usePageTitle(test?.name, "Assessment Summary");` separately and wait for `test?.name` value to change - and if falsy then we set to second param

frontend/src/components/auth/usePageTitle.ts
```diff
@@ -0,0 +1,17 @@
...
+  skip = false,
+  setOnlyIfUnchanged = false,
+) => {
+  const prevTitle = useRef("");
+  prevTitle.current = document.title;
+  useEffect(() => {
+    if (skip) return;
+    if (!title) return;
+    if (setOnlyIfUnchanged && prevTitle.current !== document.title) return;
+    document.title = `${title} | Jump Math`;
```
carissa-tang, line 7: do we need to update `prevTitle.current` when title changes? not exactly sure what this logic is doing

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -74,6 +75,13 @@ const AssessmentEditorPage = (): React.ReactElement => {
     },
   });
 
+  const assessmentName = watch("name");
+  const isExisting = !!state;
+  usePageTitle(
```
carissa-tang, line 80: i feel like it actually might make more sense if we don't change page title when test name changes - and if the user deletes the entire test name, the title becomes `"Edit Assessment"` even though they are still in editing mode
image

carissa-tang (commented):
love this !! left a few thoughts

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
jfdoming: Wait @carissa-tang wym, just display "edit assessment" permanently? I felt like it might then be hard to disambiguate between different tests but maybe not?

carissa-tang: i mean like display the original test name permanently (if in editing mode)

carissa-tang: rather than having it change as the user changes the field value

jfdoming: Oh, hmm okay. I guess that makes sense, updating on save

frontend/src/components/auth/PrivateRoute.tsx
```diff
@@ -1,33 +1,56 @@
...
+const Page = ({ component: RenderComponent, roles, title }: PageProps) => {
   const { authenticatedUser } = useContext(AuthContext);
 
-  if (!authenticatedUser) {
+  const isSignedIn = authenticatedUser !== null;
+  const isAuthorized = isSignedIn && roles.includes(authenticatedUser.role);
+
+  usePageTitle(
+    isAuthorized ? title : "Not Found",
+    !isSignedIn || typeof title === "undefined",
```
carissa-tang: why do we need both `typeof title === "undefined"` and `if (!title) return;` in `usePageTitle`? i think this check prevents "Not Found" from displaying (e.g. if teacher visits /admin).

carissa-tang (commented):
should we also display `"Not Found | Jump Math"` for a user not signed in who visits an non-existing page?

frontend/src/components/auth/PrivateRoute.tsx
jfdoming: oops, my bad. I must have missed the existing check when I was refactoring :')

carissa-tang (approved):
looks great !!

PR #532, opened 2023-08-20
fix: bugs surrounding edit-classroom modal

jfdoming:
## Notion ticket link
No ticket - bugs I noticed in the course of using the platform

## Implementation description
- Fetch missing `startDate` field
- Fix classroom page constantly rerendering
- Reset form data to default values rather than empty state
- Hydrate default values on classrooms page
- Refetch correct queries upon update

## Steps to test
1. Open the modals on the classrooms page and classroom drilldown page several times.
2. Verify that the values are reset to default (current) values each time.

* 

## Checklist

### DIFF
--- a/frontend/src/utils/GeneralUtils.ts
+++ b/frontend/src/utils/GeneralUtils.ts
@@ -1,4 +1,6 @@
 import { isSameDay } from "date-fns";
+import type { DocumentNode } from "graphql";
+import { getOperationAST } from "graphql";
 
 export const titleCase = (input: string): string => {
   const words = input.trim().split(/\s+/);
@@ -93,4 +95,12 @@ export const getLetterFromNumber = (number: number): string => {
   const secondLetter = String.fromCharCode((number % 26) + 97);
 
   return firstLetter + secondLetter;
+};
+
+export const getQueryName = (query: DocumentNode): string => {
+  const name = getOperationAST(query)?.name?.value;
+  if (!name) {
+    throw new Error("Query name not found. This is probably a bug.");
+  }
+  return name;
 };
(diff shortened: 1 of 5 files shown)


carissa-tang (commented):
could you look into this error i get when adding a new classroom?
```
"Validation failed: grades.0: `Grade 3` is not a valid enum value for path `grades.$`."
```
image

The class ends up being created, but there's an error message

jfdoming:
> could you look into this error i get when adding a new classroom?
> 
> ```
> "Validation failed: grades.0: `Grade 3` is not a valid enum value for path `grades.$`."
> ```
> 
> image
> The class ends up being created, but there's an error message

Hmm, not able to repro this... Maybe let's chat about it in the sync today

frontend/src/utils/GeneralUtils.ts
```diff
@@ -94,3 +96,11 @@ export const getLetterFromNumber = (number: number): string => {
 
   return firstLetter + secondLetter;
 };
+
+export const getQueryName = (query: DocumentNode): string => {
+  const name = getOperationAST(query)?.name?.value;
+  if (!name) {
+    throw new Error("Query name not found. This is probably a bug.");
```
carissa-tang, line 103: is there some info about the query we can add into the error message? or the query itself lol

carissa-tang (approved):
nice!

frontend/src/utils/GeneralUtils.ts
jfdoming: good point lol

edit: we decided not to bc there's nothing useful to print

PR #492, opened 2023-07-25
(2) add assessment section component for teacher dashboard

jfdoming:
## Notion ticket link

[Teacher Dashboard page](https://www.notion.so/uwblueprintexecs/Teacher-Dashboard-page-3d48a79f5d9342f5afc7f6d56fe230d5)

## Implementation description
image

- Extract hook for assessment data
- Add assessment section for dashboard page
- Support limit option on backend

## Steps to test
1. View component library page.
2. Check that things visually match the design.

## What should reviewers focus on?
* 

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #546, opened 2023-08-23
fix: test session limiting and sorting on dashboard

jfdoming:
## Notion ticket link

Follow-up to #492

## Implementation description
- Compute test session status on backend
- Limit by assessment status on backend
- Sort assessments by target date on frontend

* 

## Checklist

### DIFF
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -196,17 +196,17 @@ class TestSessionService implements ITestSessionService {
             teacher: { $eq: teacherId },
             endDate: { $gte: now },
             startDate: { $lte: now },
-          }),
+          }).sort({ endDate: 1 }),
           // upcoming
           MgTestSession.find({
             teacher: { $eq: teacherId },
             startDate: { $gt: now },
-          }),
+          }).sort({ startDate: 1 }),
           // past
           MgTestSession.find({
             teacher: { $eq: teacherId },
             endDate: { $lt: now },
-          }),
+          }).sort({ startDate: -1 }),
         ];
 
         queries.forEach((query) => query.limit(limit));
--- a/backend/services/implementations/testSessionService.ts
+++ b/backend/services/implementations/testSessionService.ts
@@ -28,8 +28,29 @@ import {
   roundTwoDecimals,
 } from "../../utilities/generalUtils";
 import calculateMarkDistribution from "../../utilities/dataVisualizationUtils";
+import { getSessionStatus } from "../../utilities/testSessionUtils";
 
 const Logger = logger(__filename);
+
+const formatTestSession =
+  (now?: Date) =>
+  (testSession: TestSession): TestSessionResponseDTO => {
+    const nowDate = now ?? new Date();
+    const sessionDTO = mapDocumentToDTO(testSession);
+
+    return {
+      ...sessionDTO,
+      status: getSessionStatus(
+        sessionDTO.startDate,
+        sessionDTO.endDate,
+        nowDate,
+      ),
+    };
+  };
+
+const formatTestSessions = (
+  testSessions: TestSession[],
+): TestSessionResponseDTO[] => testSessions.map(formatTestSession(new Date()));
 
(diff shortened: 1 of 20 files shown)


frontend/src/components/teacher/view-sessions/TestSessionListItemPopover.tsx
```diff
@@ -66,7 +66,7 @@ const TestSessionListItemPopover = ({
       onOpen={onPopoverOpen}
     >
       <VStack spacing={0}>
-        {status !== "past" && (
+        {status !== "PAST" && (
```
carissa-tang: can we make an enum to avoid these magic strings?

backend/services/implementations/testSessionService.ts
```diff
@@ -163,13 +186,42 @@ class TestSessionService implements ITestSessionService {
...
+          // upcoming
+          MgTestSession.find({
+            teacher: { $eq: teacherId },
+            startDate: { $gt: now },
+          }).sort({ startDate: 1 }),
+          // past
+          MgTestSession.find({
+            teacher: { $eq: teacherId },
+            endDate: { $lt: now },
+          }).sort({ startDate: -1 }),
```
carissa-tang, line 210: should this be endDate: -1? then you can see the most recently completed ones

backend/services/implementations/testSessionService.ts
```diff
@@ -163,13 +186,42 @@ class TestSessionService implements ITestSessionService {
     limit?: number,
   ): Promise<Array<TestSessionResponseDTO>> {
     try {
-      const query = MgTestSession.find({ teacher: { $eq: teacherId } });
+      const now = new Date();
+
+      let testSessions: TestSession[];
       if (limit !== undefined) {
-        query.limit(limit);
+        const queries = [
```
carissa-tang, line 194: would aggregation pipeline like this be better? any downsides?
image

carissa-tang (commented):
very clean. left some comments ..also perhaps add some unit tests? 👀

backend/services/implementations/testSessionService.ts
jfdoming: Unfortunately the UI shows past assessments sorted by "assigned" date rather than "closed" date (I believe this was a design decision). We could override that but I'd have to change the frontend copy. For now it has to be startDate

frontend/src/components/teacher/view-sessions/TestSessionListItemPopover.tsx
jfdoming: I feel like this didn't work when I tried it in the past but I'll have a look again

backend/services/implementations/testSessionService.ts
jfdoming: Main downside is that the sort you specific isn't valid (same key twice) and to fix it and make the right sort apply to the right objects, we'd have to extract an extra target date field here. I wanted to avoid extracting that field on the backend since it's not super clear what it means (depends on the copy we choose on the frontend), and it would make the pipeline quite large

jfdoming: Let me know your thoughts though

carissa-tang: makes sense, was curious what pipeline would look like but seems too complex

frontend/src/types/TestSessionTypes.ts
```diff
@@ -1,5 +1,13 @@
-export const STATUSES = ["active", "upcoming", "past"] as const;
-export type TestSessionStatus = (typeof STATUSES)[number];
+export const enum TestSessionStatus {
+  ACTIVE = "ACTIVE",
+  UPCOMING = "UPCOMING",
+  PAST = "PAST",
+}
+export const ALL_TEST_SESSION_STATUSES = [
```
carissa-tang: why is `ALL` necessary?

carissa-tang (approved):
would prefer if you used the enum everywhere in place of the hardcoded strings but its fine

carissa-tang (approved):

PR #564, opened 2023-09-04
fix: handle test session references on class archive (4)

carissa-tang:
## Notion ticket link

[Handle test sessions on class archive](https://www.notion.so/uwblueprintexecs/Handle-test-sessions-on-class-archive-29c372b5c77f4ee5868a877454621718?pvs=4)

## Implementation description
* On archive
   * Delete upcoming test sessions
   * Auto close active sessions → set end date to now
   * Return id (otherwise we need to refetch the updated class, and we don't need that info)
* Update archive modal to tell user that active sessions will end

## Steps to test
1. Create active and upcoming test sessions
image

2. Archive class
image

3. Check that active test session is past and upcoming test session is deleted
image

* 

## Checklist

### DIFF
--- a/backend/models/class.model.ts
+++ b/backend/models/class.model.ts
@@ -1,6 +1,5 @@
 import type { Document } from "mongoose";
 import mongoose, { Schema, model } from "mongoose";
-// eslint-disable-next-line import/no-cycle
 import { Grade } from "../types";
 
 /**
--- a/backend/models/class.model.ts
+++ b/backend/models/class.model.ts
@@ -20,8 +20,6 @@ export interface Class extends Document {
   gradeLevel: Grade;
   /** the id of the teacher that teaches the class  */
   teacher: string;
-  /** the ids of the test sessions assigned to the class */
-  testSessions: string[];
   /** the students of the class */
   students: Student[];
   /** whether the class is active or archived */
@@ -72,10 +70,6 @@ const ClassSchema: Schema = new Schema(
     teacher: {
       type: mongoose.Schema.Types.ObjectId,
       ref: "User",
-      required: true,
-    },
-    testSessions: {
-      type: [{ type: mongoose.Schema.Types.ObjectId, ref: "TestSession" }],
       required: true,
     },
     students: {
--- a/backend/models/class.model.ts
+++ b/backend/models/class.model.ts
@@ -1,9 +1,6 @@
 import type { CallbackError, Document } from "mongoose";
 import mongoose, { Schema, model } from "mongoose";
-// eslint-disable-next-line import/no-cycle
-import MgUser from "./user.model";
 import MgTestSession from "./testSession.model";
-// eslint-disable-next-line import/no-cycle
 import { Grade } from "../types";
 
 /**
@@ -20,8 +17,6 @@ export interface Class extends Document {
   gradeLevel: Grade;
   /** the id of the teacher that teaches the class  */
   teacher: string;
-  /** the ids of the test sessions assigned to the class */
-  testSessions: string[];
   /** the students of the class */
   students: Student[];
   /** whether the class is active or archived */
@@ -74,10 +69,6 @@ const ClassSchema: Schema = new Schema(
       ref: "User",
       required: true,
     },
-    testSessions: {
-      type: [{ type: mongoose.Schema.Types.ObjectId, ref: "TestSession" }],
-      required: true,
-    },
     students: {
       type: [StudentSchema],
       required: false,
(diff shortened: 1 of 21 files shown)


frontend/src/components/teacher/student-management/classroom-summary/ClassroomPopover.tsx
```diff
@@ -116,6 +116,13 @@ const ClassroomPopover = ({
         body={
           'Are you sure you want to archive this classroom? You can find your archived classrooms in the "Archived" tab.'
         }
+        customBodyText={
+          <Text color="red.200" pt={2} textStyle="paragraph">
+            Note that all <strong>upcoming</strong> assessments will be{" "}
```
carissa-tang: linter added these spaces

backend/services/implementations/classService.ts
```diff
@@ -198,6 +199,10 @@ class ClassService implements IClassService {
           `Class with id ${id} not found or not currently active`,
         );
       }
+
+      // Make a best effort to delete upcoming test sessions and end active test sessions
+      await this.deleteUpcomingTestSessions(id, date);
+      await this.endActiveTestSessions(id, date);
```
jfdoming, line 206: Maybe make a note that the `await`ed methods here catch any exceptions, which is why this is best-effort?

frontend/src/components/common/modal/Modal.tsx
```diff
@@ -38,6 +39,7 @@ const Modal = ({
   submitButtonVariant,
   cancelButtonVariant,
   onSubmit,
+  customBodyText,
```
jfdoming: I'm making a PR for supporting toasts everywhere which has to add a bunch of props to this component... could we instead use the existing `body` attribute and allow rich components, so that the props don't grow out of hand here? See my next comment

frontend/src/components/common/modal/ModalText.tsx
```diff
@@ -26,6 +28,7 @@ const ModalText = ({
           <Text color="grey.300" textStyle="paragraph">
             {body}
           </Text>
+          {customBodyText && customBodyText}
```
jfdoming: We could instead do
```suggestion
          {typeof body === "string" ? <Text color="grey.300" textStyle="paragraph">
            {body}
          </Text> : body}
```

frontend/src/components/teacher/student-management/classroom-summary/ClassroomPopover.tsx
jfdoming: happens lol

jfdoming: also is using the `strong` tag and `u` considered best practice? IIRC those tags are deprecated or something in favour of using CSS but not really sure

jfdoming (commented):
couple of thoughts but looks nice

frontend/src/components/teacher/student-management/classroom-summary/ClassroomPopover.tsx
jfdoming: To be honest, not too sure we need the emphasis here. I think red text is probably enough

carissa-tang: fair

backend/models/class.model.ts
```diff
@@ -81,4 +82,19 @@ const ClassSchema: Schema = new Schema(
...
+      // Delete all test sessions associated with class
+      await MgTestSession.deleteMany({ class: doc.id });
+    }
+  } catch (error) {
+    return next(error as CallbackError | undefined);
+  }
+
+  return next();
+});
+
```
joyce-shi: does this change -> we delete all associated test sessions when we delete a class? if so, can we update the description & link the appropriate ticket to reflect this? 

this is different from deleting test sessions associated with a class when it is archived, right?

backend/models/user.model.ts
```diff
@@ -57,18 +57,27 @@ const UserSchema: Schema = new Schema({
...
 
-  /* eslint-disable no-underscore-dangle */
-  await MgSchool.findOneAndUpdate(
-    { teachers: doc._id },
-    { $pull: { teachers: doc._id } },
-    { new: true },
-  );
-  await MgTestSession.deleteMany({ teacher: doc._id });
-  await MgClass.deleteMany({ teacher: doc._id });
+  return next();
```
joyce-shi: also here ^. have we lumped in all the delete dangling reference tickets into this PR? want to clarify the scope of these changes

backend/services/implementations/classService.ts
joyce-shi, line 206: why do we need a `date` parameter at all? can we not just get the current date in the functions themselves? or is there logic here i'm missing

joyce-shi (commented):
LGTM! just had one minor question about implementation & some questions around the scope of this PR :D 

I would approve, but I'll leave that to Julian, since it looks like he had some requested changes

frontend/src/components/common/modal/Modal.tsx
carissa-tang: good idea

backend/services/implementations/classService.ts
carissa-tang: i wanted to use the same date for both functions for consistency

backend/models/class.model.ts
carissa-tang: oops this code has already been merge - branch didn't update

backend/models/user.model.ts
carissa-tang: same as above

jfdoming (approved):
Nice

PR #565, opened 2023-09-05
(1) feat: preview question

joyce-shi:
## Notion ticket link

Preview Question

## Implementation description
* added Preview Question feature
* refactored existing student experience to be reusable for preview feature
* refactored existing student experience to use `QuestionElement` type (instead of `QuestionComponentResponse)

* No breaking changes (I am not confident that I didn't break something somewhere else in the app haha)
* Feature works as expected


frontend/src/components/student/NavButtons.tsx
```diff
@@ -37,7 +40,7 @@ const NavButtons = (): React.ReactElement => {
       )}
       <Spacer />
       {isLastQuestion ? (
-        <SubmitButton />
+        <SubmitButton isDisabled={isPreviewMode} />
```
joyce-shi, line 43: I'm not sure if the best thing to do here is to disable the submit button? Open to thoughts.

frontend/src/components/student/AssessmentExperience.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface AssessmentExperienceProps {
+  test: TestResponse | null;
+  className: string;
+}
+
+const AssessmentExperience = ({
+  test,
+  className,
+}: AssessmentExperienceProps): React.ReactElement => {
+  useReloadPrompt();
```
jfdoming: Is it intended to have this prompt both here and in `WriteAssessmentPage.tsx`? I imagine that might result in some wonky behaviour... I imagine you could remove it here since we don't need it when previewing?

frontend/src/components/admin/question-creation/PreviewQuestionModal.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useContext } from "react";
+import { Button } from "@chakra-ui/react";
+import { v4 as uuidv4 } from "uuid";
+
+import { EditOutlineIcon } from "../../../assets/icons";
+import QuestionEditorContext from "../../../contexts/QuestionEditorContext";
+import AssessmentExperience from "../../student/AssessmentExperience";
+
+const PreviewQuestionModal = (): React.ReactElement => {
```
jfdoming: This is called a `Modal`; does it actually act like one? If this is the page we discussed, maybe `PreviewQuestionPage` instead?

frontend/src/components/student/NavButtons.tsx
jfdoming: I think disabling the button makes a lot of sense!

jfdoming (commented):
This is awesome, thanks for putting in all the work for this one Joyce! Left a few comments, and maybe Carissa will want to have a look too

frontend/src/components/student/AssessmentExperience.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from "react";
```
carissa-tang, line 1: should we move this out of `student` if its reused for admins? not sure

frontend/src/components/student/StudentDashboardHeader.tsx
```diff
@@ -1,23 +1,27 @@
 import React from "react";
-import { Text, VStack } from "@chakra-ui/react";
+import { Spacer, Text, VStack } from "@chakra-ui/react";
```
carissa-tang, line 2: can we center align the header if classroomName is empty?
image

frontend/src/utils/StudentUtils.ts
```diff
@@ -7,14 +7,17 @@ import type {
 } from "../APIClients/types/TestClientTypes";
 import type { Answers } from "../types/AnswerTypes";
 import QuestionNumberTypes from "../types/QuestionNumberTypes";
+import type { Question, QuestionElement } from "../types/QuestionTypes";
```
carissa-tang, line 10: ig ideally in the future we refactor all our teacher data viz to use `QuestionElement` rather than `QuestionComponentResponse` lol

carissa-tang (commented):
lgtm! left a few thoughts

frontend/src/components/student/AssessmentExperience.tsx
joyce-shi: oops, meant to remove here :D good catch - thank you!

frontend/src/components/admin/question-creation/PreviewQuestionModal.tsx
joyce-shi: Hmm. I'm torn here.

I don't love the idea of calling it a page, since in my mind each of our pages correspond to a unique route, whereas this one doesn't. Additionally, In my opinion, it acts more like a modal since we control its display with a `showModal` state  variable that we toggle. 

However, I also see why it may not be a `Modal` in the traditional sense. Any other suggestions?

frontend/src/components/student/AssessmentExperience.tsx
joyce-shi: I had that thought, but then I feel like it makes sense to move all of its child components out of `student` as well. And then... that's like the whole `student` folder. LOL. 

I don't mind keeping it in `student` because at its core, I think the `AssessmentExperience` is basically just what the student sees, so I feel like its "student-y"? But don't feel super strongly about it.

But I'm open to ideas about renaming the folder or moving out the `AssessmentExperience` component and its children if you have suggestions on where to move it?

frontend/src/utils/StudentUtils.ts
joyce-shi: eh. I think it would make sense, but not essential. I feel like we could maybe only refactor if there's some need for it? We probably have enough on our plate as is 😅

frontend/src/components/student/StudentDashboardHeader.tsx
joyce-shi: yes ma'am

frontend/src/components/admin/question-creation/PreviewQuestionModal.tsx
jfdoming: Any reason why it doesn't get its own route? Seems like we could do some sort of /assessment/\<id\>/edit -> /assessment/\<id\>/preview (probably in a future ticket though?)

joyce-shi: Hmm, it's a less trivial change than it seems, since we don't have unique routes for the `assessment-editor` (i.e the assessment ID is not contained within the URL for the assessment editor).

Because of this, we don't have routing at all for the`question-editor`, (because if you were to go directly to `/assessment-editor/question-editor/`, what would you see? it wouldn't be attached to any assessment). So because of this problem, I implemented the sub-feature of **Preview Question** in the same fashion.

jfdoming: Yeah, I'm not saying we should do the path refactor in this pr. In fact, I have a backlog ticket I want to do once we hit our deadlines (we _will_ 🤞) to add this kind of routing to more pages. But I think calling it a "page" component would make sense here.

jfdoming:
Aside from the page comment I left above things look good but I'd have @carissa-tang take a look as well

frontend/src/components/student/AssessmentExperience.tsx
carissa-tang: makes sense, we can keep for now to avoid large refactor

frontend/src/utils/StudentUtils.ts
carissa-tang: yea fs, just a thought

carissa-tang (approved):
lgtm! 🥳

jfdoming:
Alternatively I can rename to "Page" if we do the refactor, but I do think something like "component" or "section" would be better

joyce-shi:
> Alternatively I can rename to "Page" if we do the refactor, but I do think something like "component" or "section" would be better

@jfdoming In my opinion, naming it `component` or `section` is no more precise than using `modal`. In fact, it's essentially just a full screen modal. If you're okay with leaving as is, let me know and I'll merge as is. 

If we decide to refactor some of the routing / components in the question editor experience, I'd be open to renaming components accordingly. But until then, if we rename this here, we'd have to rename other things throughout the codebase as well to be consistent and I'm not convinced of the value-add here of that effort. Also, not entirely convinced of the value-add of adding individual routes to the `assessment-editor` and `question-editor` subpages and it introduces a whole host of other edge case issues.

Let's have that conversation synchronously though!

PR #568, opened 2023-09-06
(1) feat: edit test sessions

carissa-tang:
## Notion ticket link

[Edit Test Session](https://www.notion.so/uwblueprintexecs/Edit-Test-Session-78d30bb6d7e748dda73a5005c956d984?pvs=4)

## Implementation description
* Add resolver and mutation for `updateTestSession`
* Update queries, types, and components associated with displaying an assessment row to include fields required for editing test session
* Populate data in distribute assessment flow and set page to last page on edit mode
* Add optional `isDisabled` prop to common components (`DatePicker`, `Table`)
* Update inputs ("Assessment Name", "Classroom" and "Start date") in distribute flow to be disabled if user is not editing an **upcoming** test session
   * Also remove pencil icon from Review page
* Populate class info by default when creating test session inside class 

## Steps to test
1. Edit active and upcoming test session from dashboard, assessment page, or classroom page 

https://github.com/uwblueprint/jump-math/assets/69697180/78d9c7bf-6093-4825-8935-9092991a7e0c

2. Create test session from within a class - should choose same class by default

https://github.com/uwblueprint/jump-math/assets/69697180/9ed099bf-1b0f-4e62-a71a-c2fc22145e48

* No breaking changes
* All entrypoints are considered

note: will add backend validation in follow up PR

## Checklist

### DIFF
--- a/backend/graphql/resolvers/testSessionResolvers.ts
+++ b/backend/graphql/resolvers/testSessionResolvers.ts
@@ -47,6 +47,10 @@ const testSessionResolvers = {
       _req: undefined,
       { testSession }: { testSession: TestSessionRequestDTO },
     ) => testSessionService.createTestSession(testSession),
+    updateTestSession: (
+      _req: undefined,
+      { id, testSession }: { id: string; testSession: TestSessionRequestDTO },
+    ) => testSessionService.updateTestSession(id, testSession),
     createTestSessionResult: (
       _req: undefined,
       { id, result }: { id: string; result: ResultRequestDTO },
--- a/frontend/src/components/common/table/Table.tsx
+++ b/frontend/src/components/common/table/Table.tsx
@@ -25,11 +25,13 @@ export interface TableRow<T extends Nodes = Nodes> {
 interface TableProps<T extends Nodes = Nodes> {
   headers: string[];
   rows: TableRow<T>[];
+  isDisabled?: boolean;
 }
 
 export const Table = <T extends Nodes = Nodes>({
   headers,
   rows,
+  isDisabled = false,
 }: TableProps<T>): React.ReactElement => {
   const { paginatedData, totalPages, currentPage, setCurrentPage } =
     usePaginatedData(rows);
@@ -56,8 +58,8 @@ export const Table = <T extends Nodes = Nodes>({
               <Tr
                 key={row.id || rowIndex}
                 backgroundColor={rowIndex % 2 === 0 ? "blue.50" : "grey.50"}
-                cursor="pointer"
-                onClick={row.onClick || undefined}
+                cursor={isDisabled ? "not-allowed" : "pointer"}
+                onClick={isDisabled ? undefined : row.onClick}
               >
                 {row.values.map((value, cellIndex) => (
                   <Td
--- a/frontend/src/components/common/table/Table.tsx
+++ b/frontend/src/components/common/table/Table.tsx
@@ -20,18 +20,17 @@ export interface TableRow<T extends Nodes = Nodes> {
   menu?: React.ReactElement;
   id?: string;
   onClick?: () => void;
+  isDisabled?: boolean;
 }
 
 interface TableProps<T extends Nodes = Nodes> {
   headers: string[];
   rows: TableRow<T>[];
-  isDisabled?: boolean;
 }
 
 export const Table = <T extends Nodes = Nodes>({
   headers,
   rows,
-  isDisabled = false,
 }: TableProps<T>): React.ReactElement => {
   const { paginatedData, totalPages, currentPage, setCurrentPage } =
     usePaginatedData(rows);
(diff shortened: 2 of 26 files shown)


frontend/src/components/pages/teacher/DisplayAssessmentsPage.tsx
```diff
@@ -84,7 +84,6 @@ const DisplayAssessmentsPage = (): React.ReactElement => {
                     <TestSessionListItem
                       key={session.testSessionId}
                       {...session}
-                      classroomName={session.classroomName ?? ""}
```
carissa-tang, line 87: `classroomName` already passed via `session` object

backend/graphql/resolvers/testSessionResolvers.ts
```diff
@@ -47,6 +47,10 @@ const testSessionResolvers = {
       _req: undefined,
       { testSession }: { testSession: TestSessionRequestDTO },
     ) => testSessionService.createTestSession(testSession),
+    updateTestSession: (
+      _req: undefined,
+      { id, testSession }: { id: string; testSession: TestSessionRequestDTO },
```
jfdoming, line 52: Is it possible to get the ID off of the DTO instead of passing it in separately? not sure if there's a preference

frontend/src/components/common/table/Table.tsx
```diff
@@ -25,11 +25,13 @@ export interface TableRow<T extends Nodes = Nodes> {
 interface TableProps<T extends Nodes = Nodes> {
   headers: string[];
   rows: TableRow<T>[];
+  isDisabled?: boolean;
 }
 
 export const Table = <T extends Nodes = Nodes>({
   headers,
   rows,
+  isDisabled = false,
```
jfdoming: Following the pattern we've used for rows so far, maybe we should consider adding this prop to the row instead of the table?

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
```diff
@@ -29,6 +34,11 @@ const DistributeAssessmentButton = ({
   }>(CREATE_TEST_SESSION, {
     refetchQueries: [{ query: GET_TEST_SESSIONS_BY_TEACHER_ID }],
   });
+  const [updateSession] = useMutation<{
+    updateSession: string;
+  }>(UPDATE_TEST_SESSION, {
+    refetchQueries: [{ query: GET_TEST_SESSIONS_BY_TEACHER_ID }],
```
jfdoming: nit: we should change this and the one above to use `getQueryName` so we can avoid worrying about extra parameters in the future

frontend/src/components/teacher/session-creation/DistributeAssessmentButton.tsx
```diff
@@ -50,6 +60,27 @@ const DistributeAssessmentButton = ({
...
+  const handleUpdateSession = async () => {
+    try {
+      await updateSession({
+        variables: {
+          id: testSessionId,
+          testSession,
+        },
+      });
+      history.push(DISPLAY_ASSESSMENTS_PAGE);
+      showToast({
```
jfdoming, line 73: Ah yes one more modal I forgot to refactor :')

frontend/src/components/teacher/view-sessions/TestSessionListItem.tsx
```diff
@@ -23,8 +23,13 @@ import TestSessionListItemPopover from "./TestSessionListItemPopover";
...
 export type TestSessionListItemProps = {
   testSessionId: string;
-  classroomName?: string;
+  testId: string;
   testName: string;
+  classroomId: string;
+  classroomName: string;
+  endDate: Date;
+  startDate: Date;
+  notes?: string;
```
jfdoming: There are enough props here that I'm wondering if we should just pass down `session` using the response type directly. That way we don't have to enumerate all of these props. Thoughts?

jfdoming (changes requested):
awesome 💯 ~~could we add some validation to the backend to make sure we're really only able to update the two fields users are allowed to update when it comes to active sessions? You could do a two-pass update, where the first pass runs `findOneAndUpdate` for only the `endDate` and `notes` fields with a query of `endDate` against `now` to check whether it's still active, and the second pass `findOneAndUpdate` for the remaining fields with a query of `startDate` against `now` to check whether it's active~~ LOL I should really check your other PRs first. Other comments are still relevant I think though!

backend/graphql/resolvers/testSessionResolvers.ts
carissa-tang: hm our other update functions pass id separately from RequestDTO so might make more sense to keep it that way?

jfdoming (approved):

PR #570, opened 2023-09-06
(2) feat: add backend validation for edit test session

carissa-tang:
## Notion ticket link

[Edit Test Session](https://www.notion.so/uwblueprintexecs/Edit-Test-Session-78d30bb6d7e748dda73a5005c956d984?pvs=4)

## Implementation description
* Throw error if test session is active and any of assessment, class, or start date is being changed

## Steps to test
1. yarn test

* 

## Checklist


backend/services/implementations/testSessionService.ts
```diff
@@ -338,6 +338,27 @@ class TestSessionService implements ITestSessionService {
...
     try {
+      const oldTestSession: TestSession | null = await MgTestSession.findById(
+        id,
+      );
+      if (!oldTestSession) {
+        throw new Error(`Test Session id ${id} not found`);
+      }
+
+      // If the test session is active, we don't allow teachers to update the assessment, class, or start date
+      const isActive = oldTestSession.startDate <= new Date();
```
jfdoming: nit: for testability, might be good to extract a `now` parameter similar to how we've done in the past. That way we can pass in a specific `now`.

backend/services/implementations/testSessionService.ts
```diff
@@ -338,6 +338,27 @@ class TestSessionService implements ITestSessionService {
...
+      if (
+        isActive &&
```
jfdoming: Could you also add some validation to prevent any updates after the test is over?

backend/services/implementations/testSessionService.ts
```diff
@@ -334,10 +334,41 @@ class TestSessionService implements ITestSessionService {
...
+        throw new Error(
+          `Test Session id ${id} has already ended and so cannot be updated`,
+        );
+      }
+
+      // If the test session is active, we don't allow teachers to update the assessment, class, or start date
+      const isActive = oldTestSession.startDate <= date;
+      if (
+        isActive &&
+        (oldTestSession.test.toString() !== testSession.test ||
```
jfdoming: Manually enumerating the fields here seems a bit error prone in case we had to add more fields. Maybe convert the object to JSON and use `Object.entries` to enumerate? Something like:

```
const isInvalidModification = Object.entries.some(...)
```
With exclusions in the .some for the keys we allow

backend/services/implementations/testSessionService.ts
```diff
@@ -334,10 +334,41 @@ class TestSessionService implements ITestSessionService {
...
 
     try {
+      const oldTestSession: TestSession | null = await MgTestSession.findById(
+        id,
+      );
+      if (!oldTestSession) {
+        throw new Error(`Test Session id ${id} not found`);
+      }
+
+      const date = nowDate ?? new Date();
```
jfdoming: Could you move this to the top of the function, before the query? If the backend is slow and a teacher makes a modification 1s before the deadline they might expect it to go through.

backend/services/implementations/testSessionService.ts
```diff
@@ -334,10 +334,48 @@ class TestSessionService implements ITestSessionService {
...
+      // If the test session is past, we don't allow teachers to update the test session at all
+      if (oldTestSession.endDate < date) {
+        throw new Error(
+          `Test Session id ${id} has already ended and so cannot be updated`,
+        );
+      }
+
+      // If the test session is active, we don't allow teachers to update the assessment, class, or start date
+      const isActive = oldTestSession.startDate <= date;
+      const keysToValidate = ["test", "class", "startDate"];
```
jfdoming: Oh sorry, can we do an exclusive check instead? Like a `validKeys` set and we have an error if there's a modification not in the valid keys? That way we get the benefit from my previous comment about data model updates

carissa-tang: hopefully i addressed it this time.

also side thought jw, should we be passing a partial objects on update (only fields that user intends to edit) or the whole request object

jfdoming: Agree, partial object would be ideal. But we can keep as is for now I think, nbd

jfdoming (approved):
looks like some test is failing but lgtm otherwise

PR #571, opened 2023-09-06
fix: expired refresh token issues

jfdoming:
## Notion ticket link

n/a

## Implementation description
In a previous PR, I implemented logic to automatically request a new session token if the current one has expired. That logic neglected to account for cases where the refresh token itself has been revoked, which can occur (for example) if a user resets their password. Implementation details:
- Add logging to backend to help with future debugging
- Add basic refresh-else-signout logic to failed requests

## Steps to test
This is hard to test; basically you need an expired refresh token. You can probably just take my word that this works.

* 

## Checklist

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { createRoot } from "react-dom/client";
 import { ApolloClient, ApolloProvider, InMemoryCache } from "@apollo/client";
 import { setContext } from "@apollo/client/link/context";
+import { onError } from "@apollo/client/link/error";
 import { createUploadLink } from "apollo-upload-client";
 
 import AuthAPIClient from "./APIClients/AuthAPIClient";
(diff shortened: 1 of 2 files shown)


frontend/src/index.tsx
```diff
@@ -96,10 +97,24 @@ const authLink = setContext(async (_, { headers }) => {
...
+    graphQLErrors.forEach(({ extensions }) => {
+      if (extensions.code === "UNAUTHENTICATED") {
+        foundRefreshError = true;
+      }
+    });
+
+    if (foundRefreshError) {
+      attemptRefresh().catch(console.error);
+    }
+  }
```
carissa-tang:
```suggestion
   if (graphQLErrors?.some(({ extensions }) => extensions.code === "UNAUTHENTICATED")) {
    attemptRefresh().catch(console.error);
  }
```

carissa-tang (approved):
small refactor, but looks good

PR #573, opened 2023-09-09
fix: fix question number padding

carissa-tang:
## Notion ticket link

[Fix question number padding](https://www.notion.so/uwblueprintexecs/Fix-question-number-padding-1b34e6dcd09f417a968ce14e02677ade?pvs=4)

## Implementation description
* Set max width to fix the width of the button (min width is still needed to override default button styling)

## Steps to test
1. Verify that the question boxes are the same width for all numbers
2. Verify that the width matches the Figma specs
image
image

* 

## Checklist

### DIFF
--- a/frontend/src/components/student/QuestionNumber.tsx
+++ b/frontend/src/components/student/QuestionNumber.tsx
@@ -23,13 +23,15 @@ const QuestionNumber = ({
           fontSize="20px"
           icon={<CheckmarkIcon />}
           isActive
-          minWidth="3.7rem"
+          maxWidth="3.6rem"
+          minWidth="3.6rem"
           onClick={onClick}
           variant="primary"
         />
       ) : (
         <Button
-          minWidth="3.7rem"
+          maxWidth="3.6rem"
+          minWidth="3.6rem"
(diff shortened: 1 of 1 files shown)


frontend/src/components/student/QuestionNumber.tsx
```diff
@@ -23,13 +23,15 @@ const QuestionNumber = ({
...
-          minWidth="3.7rem"
+          maxWidth="3.6rem"
+          minWidth="3.6rem"
           onClick={onClick}
           variant="primary"
         />
       ) : (
         <Button
-          minWidth="3.7rem"
+          maxWidth="3.6rem"
```
jfdoming, line 33: Just plain `width` doesn't work right

carissa-tang: yea, bc we need to override this [line](https://github.com/uwblueprint/jump-math/blob/staging/frontend/src/themes/components/button.ts#L23). alternatively, could make a new size variant but not sure its necessary

jfdoming (approved):

PR #574, opened 2023-09-10
fix: fix test session list item UI

carissa-tang:
## Notion ticket link

[Fix Assessments Card UI](https://www.notion.so/uwblueprintexecs/Fix-Assessments-Card-UI-a6f22565aabf47cbbade174833195694?pvs=4)

## Implementation description
* Remove extra padding above tab panels
* Fix bug to not show classname tag / show status tag when inside Classroom Page (bug introduced in this PR https://github.com/uwblueprint/jump-math/pull/568)
* Reduce width of classname tag (from 40 to 36)
* ~~Limit number of lines to 1 for text~~
* Refactored `TestSessionListItem` by extracting components for readability

## Steps to test
image

image

image

image

* No breaking changes

## Checklist

### DIFF
--- a/frontend/src/components/pages/teacher/DisplayClassroomPage/DisplayClassroomAssessmentsPage.tsx
+++ b/frontend/src/components/pages/teacher/DisplayClassroomPage/DisplayClassroomAssessmentsPage.tsx
@@ -76,6 +76,7 @@ const DisplayClassroomAssessmentsPage = () => {
       {paginatedData?.map((session) => (
         <TestSessionListItem
           key={session.testSessionId}
+          inClassroomPage
           isReadOnly={!data?.class.isActive}
           session={{
             ...session,
--- a/frontend/src/components/teacher/view-sessions/TestSessionListItem.tsx
+++ b/frontend/src/components/teacher/view-sessions/TestSessionListItem.tsx
@@ -11,7 +11,6 @@ import Copyable from "../../common/Copyable";
 import TestSessionListItemBody from "./TestSessionListItemBody";
 import TestSessionListItemPopover from "./TestSessionListItemPopover";
 import TestSessionListItemStatistics from "./TestSessionListItemStatistics";
-import TestSessionListItemStatusTag from "./TestSessionListItemStatusTag";
 
 export type TestSessionListItemProps = {
   session: TestSessionEditingData & {
@@ -36,6 +35,7 @@ const TestSessionListItem = ({
   const history = useHistory();
   const { testSessionId, classroomName, testName, status } = session;
   const { targetDate, accessCode, ...testSessionEditingData } = session;
+  const isPast = status === TestSessionStatus.PAST;
 
   const formattedAccessCode = `${accessCode.slice(
     0,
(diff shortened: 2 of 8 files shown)


carissa-tang:
I refactored the component bc it was pretty hard for me to read, but we can revert if you hate it lol. Also I was playing around with moving the access code to the right but not sure with how it looks in the table, so just added extra space for now. lmk which you prefer!

image
image
image

frontend/src/components/pages/teacher/DisplayClassroomPage/DisplayClassroomAssessmentsPage.tsx
```diff
@@ -76,6 +76,7 @@ const DisplayClassroomAssessmentsPage = () => {
       {paginatedData?.map((session) => (
         <TestSessionListItem
           key={session.testSessionId}
+          inClassroomPage
```
jfdoming, line 79: Wondering if we need this flag explicitly? I think before we were just using the nullish-ness of `className` for the same purpose, and having a flag providing the same information seems a little brittle

frontend/src/components/teacher/view-sessions/TestSessionListItem.tsx
```diff
@@ -64,152 +42,69 @@ const TestSessionListItem = ({
...
-        hasArrow
-        isDisabled={status !== TestSessionStatus.PAST}
-        label="Click to view statistics of this assessment and each student's performance."
-        p={2}
-        placement="bottom"
+      <HStack
+        _hover={{ bg: "grey.100" }}
+        as="button"
+        borderRadius={16}
+        gap={6}
```
jfdoming, line 70: I'm a little worried about these extra props drifting from the corresponding `HStack` below (which was why we used the `isDisabled` prop on the tooltip before). Maybe add a `WRAPPER_PROPS` constant above and spread it onto the children? lmk your thoughts though, seems like there might be a better way

frontend/src/components/teacher/view-sessions/TestSessionListItemBody.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <Text
+          color={status === TestSessionStatus.PAST ? "grey.300" : "blue.300"}
+          noOfLines={1}
+          textStyle="subtitle2"
+        >
+          {testName}
+        </Text>
+        <Text
+          color={status === TestSessionStatus.PAST ? "grey.200" : "blue.200"}
+          noOfLines={1}
```
jfdoming: Adding this prop means that the text is truncated on small screens => teachers can't always read the date in full. Could we either (a) revert the change, (b) add a tooltip for the date (seems ugly though), or (c) consider other options to make it fully visible?

jfdoming (changes requested):
I like the refactor! Left a couple of comments I think we should think about. Also, I really like how moving the code to the right side looks! However, I think with the status label it looks a bit weird. Could we actually move the status label all the way to the left along with moving the code to the right? It would take the place of the "classroom" tag since those two tags never appear at the same time here

frontend/src/components/pages/teacher/DisplayClassroomPage/DisplayClassroomAssessmentsPage.tsx
carissa-tang: `className` now exists in all pages bc it's passed down for editing

jfdoming: Wait what do we need it for when editing? I'd think we would only need classId for the currently-selected class?

carissa-tang: we display it in the review page

jfdoming: oh ok

frontend/src/components/teacher/view-sessions/TestSessionListItemBody.tsx
carissa-tang: one idea is to show untruncated on hover, but we don't really use that anywhere else so i'll just revert for now

jfdoming (approved):

PR #585, opened 2023-09-15
feat: better assessment editor loading state

jfdoming:
## Notion ticket link

[Improve save behaviour on admin assessments](https://www.notion.so/uwblueprintexecs/Improve-save-behaviour-on-admin-assessments-1dc65e756a794d4bafdab43b80bac0ff)

## Implementation description
- Add full-page whiteout instead of loading spinner
- Handle reload-prompt properly

## Steps to test
1. Open an assessment in the editor
2. Click back and verify there is no prompt
3. Open it again
4. Edit something
5. Click back and verify there is a prompt and hit cancel
6. Click save
7. Verify there is a full-screen whiteout box and you can't click any elements
8. After saving finishes, click back
9. Verify there is no prompt
10. Repeat the edit-save cycle once more
11. This time, instead of clicking back, edit a field again
12. Click back and verify you see a prompt again

* 

## Checklist

### DIFF
--- a/frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
+++ b/frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
@@ -1,5 +1,5 @@
 import React, { useContext } from "react";
-import type { SubmitHandler, UseFormHandleSubmit } from "react-hook-form";
+import { type SubmitHandler, useFormContext } from "react-hook-form";
 import { useHistory } from "react-router-dom";
 import {
   Box,
@@ -22,6 +22,7 @@ import {
 import AssessmentContext from "../../../contexts/AssessmentContext";
 import { formatDate, getCurrentDate } from "../../../utils/GeneralUtils";
 import ActionButton from "../../common/form/ActionButton";
+import useActionFormHandler from "../../common/modal/useActionFormHandler";
 import BackButton from "../../common/navigation/BackButton";
 import Popover from "../../common/popover/Popover";
 import PopoverButton from "../../common/popover/PopoverButton";
@@ -31,20 +32,18 @@ import PublishAssessmentModal from "../assessment-status/EditStatusModals/Publis
 
 interface AssessmentEditorHeaderProps {
   name: string;
-  handleSubmit: UseFormHandleSubmit<TestRequest>;
   isEditing: boolean;
   onConfirmArchive: SubmitHandler<TestRequest>;
   onConfirmPublish: SubmitHandler<TestRequest>;
   onDelete: SubmitHandler<TestRequest>;
   onSave: SubmitHandler<TestRequest>;
-  onError: () => void;
+  onError: (message: string) => void;
   validateForm: () => void;
   updatedAt?: string;
 }
 
 const AssessmentEditorHeader = ({
   name,
-  handleSubmit,
   isEditing,
   onConfirmArchive,
   onConfirmPublish,
@@ -77,11 +76,16 @@ const AssessmentEditorHeader = ({
     onPublishModalOpen();
   };
 
-  const handleSave = handleSubmit(onSave, onError);
-  const handlePublish = handleSubmit(onPublish, onError);
-  const handleConfirmPublish = handleSubmit(onConfirmPublish, onError);
-  const handleConfirmArchive = handleSubmit(onConfirmArchive, onError);
-  const handleDelete = handleSubmit(onDelete, onError);
+  const { getValues } = useFormContext<TestRequest>();
+
+  // We need validation on these actions.
+  const handleSave = useActionFormHandler(onSave);
+  const handlePublish = useActionFormHandler(onPublish);
+  const handleConfirmPublish = useActionFormHandler(onConfirmPublish);
+
+  // We don't need validation on these actions.
+  const handleConfirmArchive = async () => onConfirmArchive(getValues());
+  const handleDelete = async () => onDelete(getValues());
   const handleCloseEditor = () => history.goBack();
 
   return (
@@ -121,6 +125,7 @@ const AssessmentEditorHeader = ({
               messageOnSuccess="Assessment saved."
               minWidth="10"
               onClick={handleSave}
+              onError={onError}
               variant="secondary"
             >
               Save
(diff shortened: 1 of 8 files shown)


carissa-tang:
What about leaving an untouched question editor?

image

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -128,9 +133,7 @@ const AssessmentEditorPage = (): React.ReactElement => {
         test,
       },
     });
-
-    setCompletedForm(true);
-    history.push(ASSESSMENTS_PAGE);
+    resetDefaultValues();
```
carissa-tang: what's this line for?

jfdoming: From the commit message:
> Now that we stay on the editor page sometimes after submission, we need to fix the handling of the reload prompt (otherwise, it will never display after the form is submitted for the first time). To do this, we check if the form state has been touched (dirtied), and only allow reloading if it has not. We also have to reset the form default values as this is how `rhf` determines whether the form is dirty.

This is the "reset the form default values" part at the end!

jfdoming:
> What about leaving an untouched question editor?
> 
> image

hmm, I felt like that wasn't a concern of this ticket, but I'll fix it for this PR, sure

carissa-tang (commented):
unrelated to this PR but save error shouldn't show here

image

also when i create a new test, there shouldn't be the warning (+ should we also not redirect users to assessments page after create?)

image

jfdoming:
> > What about leaving an untouched question editor?
> > image
> 
> hmm, I felt like that wasn't a concern of this ticket, but I'll fix it for this PR, sure

actually looks like this will be fairly complex, I'd like to make this a separate ticket

jfdoming:
separate ticket: https://www.notion.so/uwblueprintexecs/Fix-reload-prompt-for-question-editor-dd2c97a68b2a44de91462b8b411569e9

frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
```diff
@@ -134,14 +134,14 @@ const AssessmentEditorHeader = ({
...
-                {isEditing && (
+            {isEditing && (
+              <Popover>
+                <VStack divider={<Divider />} spacing="0em">
                   <PopoverButton name="Archive" onClick={onArchiveModalOpen} />
-                )}
-                <PopoverButton name="Delete" onClick={onDeleteModalOpen} />
-              </VStack>
-            </Popover>
+                  <PopoverButton name="Delete" onClick={onDeleteModalOpen} />
```
carissa-tang, line 144: i think design originally wanted delete option, in which if its a new assessment the editor just closes. i'm fine w getting rid of it - but can we clean up [this line](https://github.com/uwblueprint/jump-math/blob/staging/frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx#L159) then? and remove `handleCloseEditor`

carissa-tang (commented):
now when i click save without filling out everything, i see "Assessment saved" lol

image

and just to confirm, we're intentionally leaving the window pop up after clicking `Create Assessment` > `Back`?

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -129,16 +152,15 @@ const AssessmentEditorPage = (): React.ReactElement => {
       },
     });
 
-    setCompletedForm(true);
-    history.push(ASSESSMENTS_PAGE);
+    onAfterModify(options);
```
carissa-tang: can we just set `redirect: true` here? feel like the extra code is unnecessary

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -105,18 +110,36 @@ const AssessmentEditorPage = (): React.ReactElement => {
     }
   };
 
+  const onAfterModify = (options?: { redirect?: boolean }) => {
+    // Mark the form as clean after updating.
+    resetForm(getValues());
+    if (options?.redirect ?? true) {
+      console.log("setting");
```
carissa-tang:
```suggestion
```

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -105,18 +110,36 @@ const AssessmentEditorPage = (): React.ReactElement => {
     }
   };
 
+  const onAfterModify = (options?: { redirect?: boolean }) => {
+    // Mark the form as clean after updating.
+    resetForm(getValues());
```
carissa-tang: we only need to reset form if we aren't redirecting right? or am i misunderstanding

frontend/src/components/common/navigation/useReloadPrompt.ts
```diff
@@ -1,19 +1,21 @@
...
-const beforeUnloadListener = (event: BeforeUnloadEvent) => {
-  event.preventDefault();
-  return (event.returnValue = "");
-};
-
-const useReloadPrompt = () => {
+const useReloadPrompt = (when?: boolean) => {
   useEffect(() => {
+    const beforeUnloadListener = (event: BeforeUnloadEvent) => {
+      if (!when) return;
```
carissa-tang: does it work if you do smth like `if (!when) return () => {};`? rather than adding the `when` logic in the event listeners

frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
```diff
@@ -77,12 +74,16 @@ const AssessmentEditorHeader = ({
...
-  const handleConfirmPublish = handleSubmit(onConfirmPublish, onError);
-  const handleConfirmArchive = handleSubmit(onConfirmArchive, onError);
-  const handleDelete = handleSubmit(onDelete, onError);
-  const handleCloseEditor = () => history.goBack();
+  const { getValues } = useFormContext<TestRequest>();
+
+  // We need validation on these actions.
+  const handleSave = useActionFormHandler(onSave);
+  const handlePublish = useActionFormHandler(onPublish);
+  const handleConfirmPublish = useActionFormHandler(onConfirmPublish);
```
carissa-tang: do we need to validate again for confirmPublish if we already did for publish?

carissa-tang (approved):

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
```diff
@@ -60,25 +68,34 @@ const AssessmentEditorPage = (): React.ReactElement => {
...
+        assessmentType: state.assessmentType,
+        status: state.status,
+        curriculumCountry: state.curriculumCountry,
+        curriculumRegion: state.curriculumRegion,
+      });
+      setQuestions(state.questions);
+    }
+  }, [state, resetForm]);
+  const isDirty =
+    !redirectTo && (isFormDirty || questions !== state?.questions);
```
carissa-tang, line 98: btw i think this doesn't work for a new test bc `questions = []` but `state?.questions = undefined`. could we handle that case?

jfdoming: @carissa-tang I thought that for new tests we said we would always display the reload prompt?

frontend/src/components/common/navigation/useReloadPrompt.ts
jfdoming: Yeah, not much difference practically but I can update

frontend/src/components/admin/assessment-creation/AssessmentEditorHeader.tsx
jfdoming: True

frontend/src/components/pages/admin/AssessmentEditorPage.tsx
carissa-tang: oh ok sure

PR #589, opened 2023-09-16
fix: popover styling

joyce-shi:
## Notion ticket link

[Fix popover styling](https://www.notion.so/uwblueprintexecs/Fix-popover-styling-4d1ca583ee0c42949225d32162b34299?pvs=4)

## Implementation description
* Updated `RemoveUser` popover to be consistent with other popovers
* Added hover behaviour to the popover buttons
* Added drop shadow to popover

## Notes
I'm still not entirely happy with how this popover looks. However, in the grand scheme of things I don't think it's that important.

### DIFF
--- a/frontend/src/components/common/popover/Popover.tsx
+++ b/frontend/src/components/common/popover/Popover.tsx
@@ -40,6 +40,7 @@ const Popover = ({
       onClose={onClose}
       onOpen={onOpen}
       placement={placement}
+      size="status"
     >
       <PopoverTrigger>{triggerButton}</PopoverTrigger>
       <PopoverContent
@@ -51,7 +52,7 @@ const Popover = ({
         overflow="hidden"
         width="100%"
       >
-        <PopoverBody padding="0 !important">{children}</PopoverBody>
+        <PopoverBody>{children}</PopoverBody>
       </PopoverContent>
     </ChakraPopover>
   );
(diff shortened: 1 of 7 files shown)


frontend/src/components/common/popover/Popover.tsx
```diff
@@ -46,11 +46,12 @@ const Popover = ({
         backgroundColor="grey.100"
         border="none"
         borderRadius="16px"
+        boxShadow="4px 4px 40px 5px rgba(0, 0, 0, 0.08)"
         maxHeight="50%"
         overflow="hidden"
         width="100%"
       >
-        <PopoverBody>{children}</PopoverBody>
+        <PopoverBody padding="0 !important">{children}</PopoverBody>
```
jfdoming: just to double check, assuming this doesn't work without the `!important`, even if we add it to the theme as a multipart style?

jfdoming (approved):
awesome! 💯 

one thing I noticed is that when there's a lot of white around the drop shadow/colour combination makes things really blend into the background. ex:
image

wondering if choosing a slightly darker drop-shadow might help?

joyce-shi:
> awesome! 💯
> 
> one thing I noticed is that when there's a lot of white around the drop shadow/colour combination makes things really blend into the background. ex: image
> 
> wondering if choosing a slightly darker drop-shadow might help?

just going by the drop shadow specified in the figma. feel free to take up the discussion with design and let me know if you're able to arrive at a decision. 👍

frontend/src/components/common/popover/Popover.tsx
joyce-shi: good catch. updated.

PR #590, opened 2023-09-16
fix: loading state for buttons

joyce-shi:
## Notion ticket link

[Add loading state to buttons](https://www.notion.so/uwblueprintexecs/Add-loading-state-to-buttons-29db3f493a4d412084eddd60af59a107?pvs=4)

## Implementation description
* Loading states for login, logout and authentication buttons

* No missed cases. I didn't add loading states for navigation buttons (i.e `history.push()`), since I don't think there is a significant enough loading time to add the loading state to the button. Additionally, the page being navigated to should have a loading state.
Note: I skipped `Modal` and `ActionButton` cases, since I figured they have already been covered in previous PRs. 👍


frontend/src/components/common/modal/ModalFooterButtons.tsx
```diff
@@ -1,32 +0,0 @@
...
-        Discard
-      </Button>
-      <Button minWidth="108px" onClick={onSave} variant="primary" width="108px">
-        Save
-      </Button>
-    </>
-  );
-};
-
-export default ModalFooterButtons;
```
joyce-shi, line 32: It looked like this file was unused, so I deleted it. Let me know if I am missing context here!

frontend/src/components/auth/teacher-signup/NavigationButtons.tsx
```diff
@@ -10,16 +10,23 @@ interface NavigationButtonsProps {
...
   onBackClick,
   continueButtonText = "Continue",
   backButtonText = "Back",
+  continueIsLoading = false,
 }: NavigationButtonsProps): React.ReactElement => {
   return (
     <>
-      <Button onClick={onContinueClick} variant="primary" width="100%">
+      <Button
+        isLoading={continueIsLoading}
```
jfdoming: nit: I'm fine with not using `ActionButton` in most cases since it won't save much in terms of readability, but in this case, using an Action button would eliminate the need for adding this flag to the props! I'd suggest that refactor if you have time

joyce-shi: good call. i'll refactor.

jfdoming (approved):

PR #599, opened 2023-09-30
feat: add routes to assessment editor

jfdoming:
## Notion ticket link

[Refactor create-assessment pages to use routes for each page](https://www.notion.so/uwblueprintexecs/Refactor-create-assessment-pages-to-use-routes-for-each-page-0a573129441a478ab75520d4fbbc2947)

## Implementation description
* Long but it works :') This feels like the home stretch somehow

Please see my question below!

Note: reviewing by commit is probably a bad idea. I was unable to make nice commits for this one, sorry 😞 

## Steps to test
1. Play around with creating and editing assessments, and verify you can access all pages without 404 errors. Try reloading the page too!

* 

## Checklist


jfdoming:
One note: I have this object called `redirectableHistory` right now, which is essentially just a `history` object that allows you to navigate without triggering the page `Prompt`. (Normal react-router `history` objects _do_ trigger it.) The best name I could come with is `redirectableHistory`, but I'm definitely open to any suggestions on better/clearer names!

frontend/src/constants/Routes.ts
```diff
@@ -9,8 +9,43 @@ export const EMAIL_ACTION_PAGE = "/email-action";
...
+export const ASSESSMENT_EDITOR_QUESTION_EDITOR_PAGE = ({
+  assessmentId,
+  questionIndex,
+}: {
+  assessmentId?: string;
+  questionIndex?: string;
+}) =>
+  ASSESSMENT_EDITOR_BASE({ assessmentId }) +
+  "/questions/" +
+  (questionIndex ? `${questionIndex}/edit` : "add");
```
carissa-tang: `add` seems a bit weird to me. would `new` be better?

frontend/src/components/admin/assessment-creation/AssessmentPreview.tsx
```diff
@@ -1,18 +1,32 @@
...
+}: AssessmentPreviewProps): ReactNode => {
+  useEffect(() => {
+    if (!questions.length) {
+      goBack();
+    }
+  }, [goBack, questions]);
+
+  if (!questions.length) {
+    return null;
+  }
```
carissa-tang, line 26: is this necessary? since we already validate form before preview so questions will never be empty

frontend/src/contexts/AssessmentContext.ts
```diff
@@ -1,31 +1,38 @@
...
+import type { SetStateAction } from "react";
 import { createContext } from "react";
+import type { LocationDescriptor } from "history";
 
+import type { RedirectableNavigatePromptHistory } from "../components/common/navigation/useRedirectableNavigatePrompt";
 import type { Question } from "../types/QuestionTypes";
 
 type AssessmentContextType = {
+  redirectableHistory: RedirectableNavigatePromptHistory;
+  setQuestionEditorDirty: (_isDirty: SetStateAction<boolean | null>) => void;
```
carissa-tang: is there a way to set this in `QuestionEditorContext`? and handle the reload prompt logic separately in QuestionEditor

frontend/src/components/admin/assessment-creation/AssessmentQuestions.tsx
```diff
@@ -23,7 +28,14 @@ const AssessmentQuestions = (): React.ReactElement => {
         <Spacer />
         <Button
           leftIcon={<PlusOutlineIcon />}
-          onClick={() => setShowQuestionEditor(true)}
+          onClick={() => {
+            setQuestionEditorDirty(false);
```
carissa-tang: i don't love that we set this in every file. is there a way to do this within QuestionEditor?

frontend/src/components/admin/question-creation/QuestionEditor.tsx
```diff
@@ -42,23 +60,43 @@ const QuestionEditor = (): React.ReactElement => {
...
     >
-      {showQuestionPreview ? (
-        <QuestionPreview />
-      ) : (
-        <>
+      <Switch>
+        <Route
+          exact
+          path={Routes.ASSESSMENT_EDITOR_QUESTION_EDITOR_PAGE({
+            assessmentId: assessmentId && ":assessmentId",
```
carissa-tang, line 77: can these params be set within the `Routes` file?

carissa-tang (commented):
some questions after first pass

frontend/src/components/admin/assessment-creation/AssessmentPreview.tsx
jfdoming: Yeah, the page is now loadable directly by URL so we need validation in that case (e.g., invalid assessment ID)

frontend/src/contexts/AssessmentContext.ts
jfdoming: I tried for a while and couldn't get it to work....  the problem is that in the component tree, the question editor technically lives underneath the assessment editor, so we have to have some way of disabling the top-level prompt. The alternative would be a refactor to move the question editor into its own component tree, and the effort there doesn't seem worth it to me.

frontend/src/constants/Routes.ts
jfdoming: Makes no difference to me, I can change it

frontend/src/components/admin/assessment-creation/AssessmentQuestions.tsx
jfdoming: I wasn't able to get this to work consistently. I'll give it another shot tho

frontend/src/components/admin/question-creation/QuestionEditor.tsx
jfdoming: Unfortunately, since we have 3 cases here (create assessment, edit assessment, and render route path), there's no way to do that without making some really ugly code. For example, I tried putting this logic inside the routes file by using `null` for "create" and `undefined` for route paths, but that got super confusing and non-obvious so I reverted it. Unless you think there's huge value in moving it to that file I'd prefer to keep it this way

frontend/src/components/admin/assessment-creation/AssessmentPreview.tsx
carissa-tang: makes sense

frontend/src/components/admin/assessment-creation/AssessmentQuestions.tsx
jfdoming: Got this to work!

frontend/src/contexts/AssessmentContext.ts
jfdoming: There's another problem I forgot about here, which is that `react-dom` prevents you from even creating two `<Prompt>`s at once!

frontend/src/components/admin/question-creation/QuestionEditor.tsx
```diff
@@ -12,26 +21,46 @@ import QuestionPreview from "./QuestionPreview";
...
-  const [showQuestionPreview, setShowQuestionPreview] = React.useState(false);
+    useState(false);
+  const [showAddMultiSelectModal, setShowAddMultiSelectModal] = useState(false);
+  const [showAddFractionModal, setShowAddFractionModal] = useState(false);
+  const [showEditorError, setShowEditorError] = useState(false);
+
+  useEffect(() => {
+    // on page load
+    setQuestionEditorDirty(false);
+  }, [editorQuestion.elements, setQuestionEditorDirty]);
```
carissa-tang:
```suggestion
  }, [editorQuestion?.elements, setQuestionEditorDirty]);
```

frontend/src/components/admin/question-creation/question-elements/ImageElement.tsx
```diff
@@ -60,7 +60,7 @@ const ImageElement = ({ id, data }: ImageElementProps): React.ReactElement => {
         prevElements,
         error,
       );
-    });
+    }, false);
```
carissa-tang: why is this false? image edit should still show the prompt no?

frontend/src/components/admin/question-creation/QuestionEditor.tsx
jfdoming: Oops, confusing that TS doesn't convert `array[naNAbleNumber]` to be `array[number] | undefined`

frontend/src/components/admin/question-creation/question-elements/ImageElement.tsx
jfdoming: Oh good point. I was trying to fix an issue where the image would immediately be marked as dirty upon rendering, but I think I need to rework this a bit.

frontend/src/components/admin/question-creation/question-elements/ImageElement.tsx
```diff
@@ -36,10 +36,14 @@ const ImageElement = ({ id, data }: ImageElementProps): React.ReactElement => {
...
     fileReader.onload = (e) => {
       if (e.target?.result) {
-        setImageMetadataRequest({
+        const metadata = {
           previewUrl: e.target.result as string,
           file,
-        });
+        };
+        setImageMetadataRequest(metadata);
+        setQuestionElements((prevElements) =>
```
carissa-tang, line 44: nice clean up

jfdoming:
image

PR #604, opened 2023-10-24
feat: preview assessment from assessments page

joyce-shi:
## Notion ticket link

[Preview Assessment button in popover menu](https://www.notion.so/uwblueprintexecs/Preview-Assessment-button-in-popover-menu-b0dd3b21848f4c11a630183938d2d7a2?pvs=4)

## Implementation description
* Preview assessment from popover button (do we want to be able to click on the row to preview the assessment? we can talk about this during Sunday work session)
* Route from assessment editor to separate preview assessment page so that it can be reused


frontend/src/constants/Routes.ts
```diff
@@ -55,6 +50,12 @@ export const ASSESSMENT_EDITOR_QUESTION_PREVIEW_PAGE = ({
   ASSESSMENT_EDITOR_QUESTION_EDITOR_BASE({ assessmentId, questionIndex }) +
   "/preview";
 
+export const ASSESSMENT_PREVIEW_PAGE = ({
+  assessmentId,
+}: {
+  assessmentId?: string;
+}) => "/admin/assessment/" + assessmentId + "/preview";
```
joyce-shi, line 57: I think it would be kind of nice to overload the route here... i.e `/admin/assessments/ + assessmentId + /preview`, but I can't figure out why that isn't working & doesn't seem like a big enough issue to be a blocker... let me know if you have thoughts (:

jfdoming: Yeah routing is tricky for this page... Happy to pair at some point if that would help. Or I can look into what's happening later this week

joyce-shi: Yah, we can pair on Sunday? This PR isn't a blocker so I don't mind leaving it open

frontend/src/components/admin/assessment-status/EditStatusButtons/PreviewButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { data } = useQuery<{ test: TestResponse }>(GET_TEST, {
+    fetchPolicy: "cache-and-network",
+    variables: { id: assessmentId },
+  });
+  const { showToast } = useToast();
+
+  return (
+    <PopoverButton
+      name="Preview"
+      onClick={async () => {
```
jfdoming: Any reason why this function is async?

frontend/src/components/pages/admin/AssessmentPreviewPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+    () =>
+      test && {
+        ...test,
+        questions: formatQuestionsResponse(test.questions),
+      },
+    [test],
+  );
+
+  const assessmentName = state?.name;
+  usePageTitle(`Previewing "${assessmentName}`);
```
jfdoming: Not sure if this was already in the logic before, but this will display "Previewing undefined" while the test is loading. Could you add a fallback?

frontend/src/components/pages/admin/AssessmentEditorPage/AssessmentEditor.tsx
```diff
@@ -272,22 +271,6 @@ const AssessmentEditor = ({ state }: AssessmentEditorProps): ReactElement => {
                   </VStack>
                 </VStack>
               </Route>
-              <Route
```
jfdoming, line 275: Wondering why this change to remove the routes here was necessary? I think this might be the cause of the routing woes from your other comment. In particular, my understanding is that the `AssessmentEditor` component "owns" the route `/admin/assessments/`, which makes it capture all sub-routes. We could potentially rename the "owner" component to something more generic?

frontend/src/components/admin/assessment-status/EditStatusButtons/PreviewButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface PreviewButtonProps {
+  assessmentId: string;
+}
+
+const PreviewButton = ({
+  assessmentId,
+}: PreviewButtonProps): React.ReactElement => {
+  const history = useHistory();
+  const { data } = useQuery<{ test: TestResponse }>(GET_TEST, {
```
jfdoming, line 19: Iirc the tests table already loads this data for each row. Is there any way you could reuse that instead? Otherwise each user will be hammering the backend twice for each test that exists in the database

frontend/src/components/pages/admin/AssessmentEditorPage/AssessmentEditor.tsx
joyce-shi: @jfdoming I can't recall exactly, but we resolved this discussion synchronously right?

frontend/src/constants/Routes.ts
joyce-shi: Same here, we resolved this synchronously right? 😅

frontend/src/components/admin/assessment-status/EditStatusButtons/PreviewButton.tsx
joyce-shi: Hmmm, but does it? Because the table runs the `GET_ALL_TESTS` query, right? Which is all the test metadata without the specific questions.

Here, we call `GET_TEST` which queries for the test metadata + questions for one _specific_ test. I think this is similar to the implementation of `EditButton`.

Let me know if I am misunderstanding your comment here!

frontend/src/components/pages/admin/AssessmentPreviewPage.tsx
joyce-shi: Good idea, done!

frontend/src/components/admin/assessment-status/EditStatusButtons/PreviewButton.tsx
jfdoming: Right. I'm wondering if it would be better to just add the additional data to the GET_ALL_TESTS query? Since imho our concern here might not be bandwidth but compute scale, since we have a single-replica base tier backend

jfdoming: I guess I'm also thinking that EditButton could use a different implementation

frontend/src/components/pages/admin/AssessmentEditorPage/AssessmentEditor.tsx
jfdoming: I know something wasn't working that did require a code change. Did you commit that change?

PR #607, opened 2023-10-29
feat: add times to assessment date pickers

jfdoming:
## Notion ticket link

[Add start / end time in test session](https://www.notion.so/uwblueprintexecs/Add-start-end-time-in-test-session-b09c7292d57c43a0a9fd271a82f5dd02)

## Implementation description
times! Add times to assessment date pickers. Classroom picker to follow.

## Steps to test
1. Create a new assessment (as a teacher) and set a time.
2. Edit an assessment and view/set the time.

* 

## Checklist

### DIFF
--- a/frontend/src/components/common/form/DateTimePicker.tsx
+++ b/frontend/src/components/common/form/DateTimePicker.tsx
@@ -33,12 +33,12 @@ const DateTimePicker = ({
       onChange(combineDateAndTime(newDate, timeValue));
     }
 
-    // For some reason, the time picker doesn't focus properly if we click it
-    // while the date picker is open. This is a workaround to make sure the
-    // click happens after the date picker has started to close.
+    // For some reason, the time picker doesn't focus properly if we try to
+    // focus it while the date picker is open. This is a workaround to make
+    // sure the focus happens after the date picker has started to close.
     setTimeout(() => {
       timePickerRef.current?.click();
-    }, 0);
+    });
   };
 
   const handleTimeChange = (newTime: Date | null) => {
--- a/frontend/src/components/common/form/DateTimePicker.tsx
+++ b/frontend/src/components/common/form/DateTimePicker.tsx
@@ -1,0 +1,70 @@
+import type { ReactElement } from "react";
+import React, { useRef, useState } from "react";
+import { HStack } from "@chakra-ui/react";
+
+import { combineDateAndTime } from "../../../utils/DateUtils";
+
+import DatePicker from "./DatePicker";
+import TimePicker from "./TimePicker";
+
+type DateTimePickerProps = {
+  name?: string;
+  onChange: (date: Date | null) => void;
+  value: Date | null | undefined;
+  isDisabled?: boolean;
+};
+
+const DateTimePicker = ({
+  isDisabled,
+  onChange,
+  value,
+  name,
+}: DateTimePickerProps): ReactElement => {
+  const timePickerRef = useRef<HTMLDivElement>(null);
+
+  const [internalDateValue, setInternalDateValue] = useState<Date | null>(null);
+  const [internalTimeValue, setInternalTimeValue] = useState<Date | null>(null);
+  const dateValue = value ?? internalDateValue;
+  const timeValue = value ?? internalTimeValue;
+
+  const handleDateChange = (newDate: Date | null) => {
+    setInternalDateValue(newDate);
+    if (timeValue && newDate) {
+      onChange(combineDateAndTime(newDate, timeValue));
+    }
+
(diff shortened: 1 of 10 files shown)


frontend/src/components/common/form/DateTimePicker.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const dateValue = value ?? internalDateValue;
+  const timeValue = value ?? internalTimeValue;
+
+  const handleDateChange = (newDate: Date | null) => {
+    setInternalDateValue(newDate);
+    if (timeValue && newDate) {
+      onChange(combineDateAndTime(newDate, timeValue));
+    }
+
+    // For some reason, the time picker doesn't focus properly if we click it
```
carissa-tang: what does this mean exactly?

carissa-tang: also, when the date picker is open and i click the hours and then the minutes, both stay open. nbd bc you can click out of it but curious if this is related

image

frontend/src/components/common/form/TimePicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+};
+
+const TimePicker = forwardRef(function TimePicker(
+  {
+    name,
+    value,
+    defaultDay,
+    onChange,
+    isDisabled,
+    attachmentDirection,
```
carissa-tang: are we using this `attachmentDirection`?

frontend/src/components/common/form/TimePicker.tsx
```diff
@@ -0,0 +1,216 @@
+import type { ForwardedRef, ReactElement } from "react";
```
carissa-tang, line 1: what do you think abt center aligning instead of aligning ends

image

image

and decreasing padding at the ends from 4->2
image

carissa-tang: also idk if we necessarily need the blue background. i feel like the cursor could be clear enough. thoughts?
image

frontend/src/components/common/form/TimePicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+    ...provided,
+    display: "none",
+  }),
+  container: (provided) => ({
+    ...provided,
+    flex: 1,
+  }),
+  menu: (provided) => ({
+    ...provided,
+    w: 200,
```
carissa-tang:
```suggestion
    w: 100,
```

carissa-tang (commented):
should we update these to show start time (for upcoming) and end time (for active)?

image

jfdoming:
> should we update these to show start time (for upcoming) and end time (for active)?
> 
> image

I think in the interest of small screens, since this is just the start date, I think it would be fine to truncate to the nearest day? Let me know if you see a justification for including the exact time, but IMHO if a teacher assigns multiple assessments on the same day they'll probably name them like "Counting Test," "Shapes Test," etc., so it won't be hard to differentiate.

frontend/src/components/common/form/TimePicker.tsx
jfdoming: Good point! This was left over from when I was playing around with designs, and I left it in in case we wanted to go back. I'll remove it now!

frontend/src/components/common/form/TimePicker.tsx
jfdoming: I have a strong preference for end alignment because (1) the cursor is placed in a more logical location, and (2) it looks more like a timestamp that you've entered in text, i.e., `03:05` instead of `03 : 05`. Let me know if you see a good reason otherwise, though.

In terms of the blue background, I included it because this matches the styling of other time pickers, including a library from Adobe and the built-in picker in Chrome. I'm not opposed to changing it, though I do like how the blue background is an indicator on which field is focused (aside from the caret, which might be hard to see for some people on its own). What do you think?

frontend/src/components/common/form/DateTimePicker.tsx
jfdoming: Nice catch with the bug! It's a similar issue but not strictly related. Essentially, the problem is that the date picker component we chose to use steals focus when it starts to close, so we need to make sure we only request focus in other components AFTER it has already started closing. The comment above is to fix the problem when you TAB into the time picker; I added a similar fix for when you click it directly, which is the bug you found.

frontend/src/components/common/form/TimePicker.tsx
carissa-tang: we can keep as is

carissa-tang (approved):
awesome work !! 🚀

PR #609, opened 2023-10-30
feat: add logic for image reference counting

carissa-tang:
## Notion ticket link

[Add Reference Counting to Images](https://www.notion.so/uwblueprintexecs/Add-Reference-Counting-to-Images-b20aa3d9a1694844945e4de8180f23b3?pvs=4)

## Implementation description
* Add `ImageCount` model, interface, implementation, and unit tests
* Update ImageUploadService
   * Initialize count on image creation 
   * Decrement count on delete, and delete image only if count is 0
   * Add `incrementImageCount` function for test service
   * Increment count on upload of existing image (i.e. filePath exists)
* Update TestService
   * Increment image count on duplicate test

## Steps to test
1. Create a test with an image. Check that the reference count is 1 and the image is uploaded to GCP.
   image

2. Duplicate a test. Check that the reference count is 2. Check that the same image is still in GCP and there is no new upload.
   image

3. Delete the duplicate test. Check that the reference count is 1. Check that the same image is still in GCP.
   image

4. Delete the original test. Check that the record is deleted from the DB. Check that the image is deleted from GCP.

* 

## Checklist

### DIFF
--- a/backend/services/implementations/testService.ts
+++ b/backend/services/implementations/testService.ts
@@ -271,29 +271,14 @@ class TestService implements ITestService {
       if (!test) {
         throw new Error(`Test ID ${id} not found`);
       }
+      // eslint-disable-next-line no-underscore-dangle
+      test._id = new mongoose.Types.ObjectId();
+      test.name += " [COPY]";
+      test.isNew = true;
+      test.status = AssessmentStatus.DRAFT;
+      test.save();
 
       await this.incrementImageCounts(test.questions);
-      try {
-        // eslint-disable-next-line no-underscore-dangle
-        test._id = new mongoose.Types.ObjectId();
-        test.name += " [COPY]";
-        test.isNew = true;
-        test.status = AssessmentStatus.DRAFT;
-        test.save();
-      } catch (mongoDbError) {
-        // rollback image upload in GCP
-        try {
-          await this.deleteImages(test.questions);
-        } catch (imageError) {
-          Logger.error(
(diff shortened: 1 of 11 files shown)


backend/services/implementations/testService.ts
```diff
@@ -272,6 +277,8 @@ class TestService implements ITestService {
       test.isNew = true;
       test.status = AssessmentStatus.DRAFT;
       test.save();
+
+      await this.incrementImageCounts(test.questions);
```
jfdoming: I think it might be better to invert the order of these two lines. If `save` passes but `increment` fails, then we might end up deleting an image before it's actually unused which is bad UX. Otherwise, we might end up retaining images for longer than we need them, but I think this is more reasonable as we can always manually delete mismatched images.

Also, shouldn't we be `await`ing on `test.save()`? I think right now the endpoint returns before the test is actually registered in the database. nbd before, but now that the logic order matters I think we ought to await

backend/services/implementations/testService.ts
```diff
@@ -400,6 +407,15 @@ class TestService implements ITestService {
     );
   }
 
+  private async incrementImageCounts(
+    questions: QuestionComponent[][],
+  ): Promise<void> {
+    await this.processImages<ImageMetadata>(
+      questions,
+      this.imageUploadService.incrementImageCount.bind(this.imageUploadService),
+    );
```
jfdoming, line 417: Since this runs for all images, have you done any sort of load test before/after? There's probably minimal difference considering things should theoretically run in parallel, but it would be good to verify

